PR #3, opened 2021-03-11
loginAccessId

renanaAkeyless:


d-goro (approved):

charts/akeyless-api-gateway/Chart.yaml
```diff
@@ -14,7 +14,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.0.2
+version: 1.0.3
```
liorhasson, line 17: if its a feature, bump it as minor

PR #10, opened 2021-04-07
asm-1462 added policies json validation

niv-ez:
Added a Helm post-install pod for validating policies JSON structure (from values.yaml)

### DIFF
--- a/charts/akeyless-zero-trust-web-access/values.yaml
+++ b/charts/akeyless-zero-trust-web-access/values.yaml
@@ -177,6 +177,7 @@ webWorker:
         }
 
 dlp:
+  enabled: false
   config:
     hostAddress: ""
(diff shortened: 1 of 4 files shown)


charts/akeyless-zero-trust-web-access/values.yaml
```diff
@@ -177,7 +177,6 @@ webWorker:
         }
 
 dlp:
-  enabled: false
```
liorhasson, line 180: revert

liorhasson (changes requested):

charts/akeyless-zero-trust-web-access/values.yaml
niv-ez: fixed

liorhasson (approved):

PR #35, opened 2021-08-01
bugfix/ingress_secret_tls - ingress tls secret use wrong hostname

liorhasson:
Ingress tls secret was issued wrong hostname.

Expected:
if you added the following hostnames under ingress:
```
- ui.gateway.local
- hvp.gateway.local
```

the chart will generate the following:
```
tls:
    - hosts:
      - ui.gateway.local
      secretName: ui.gateway.local-tls
    - hosts:
      - hvp.gateway.local
      secretName: hvp.gateway.local-tls
```

### DIFF
--- a/charts/akeyless-api-gateway/templates/_helpers.tpl
+++ b/charts/akeyless-api-gateway/templates/_helpers.tpl
@@ -66,10 +66,10 @@ Create the name of the service account to use
 Get the Ingress TLS secret.
 */}}
 {{- define "akeyless-api-gw.ingressSecretTLSName" -}}
-    {{- if .Values.ingress.existingSecret -}}
-        {{- printf "%s" .Values.ingress.existingSecret -}}
+    {{- if .existingSecret -}}
+        {{- printf "%s" .existingSecret -}}
     {{- else -}}
-        {{- printf "%s-tls" .Values.ingress.hostname -}}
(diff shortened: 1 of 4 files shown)


charts/akeyless-api-gateway/templates/deployment.yaml
```diff
@@ -58,13 +58,13 @@ spec:
             {{- end }}
           livenessProbe:
             httpGet:
-              path: /status
-              port: 18888
+              path: /
+              port: 8080
```
liorhasson, line 62: more reliable then UI port (18888)

devorbitus (approved):
LGTM

charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -66,10 +66,10 @@ Create the name of the service account to use
 Get the Ingress TLS secret.
 */}}
 {{- define "akeyless-api-gw.ingressSecretTLSName" -}}
-    {{- if .Values.ingress.existingSecret -}}
-        {{- printf "%s" .Values.ingress.existingSecret -}}
+    {{- if .existingSecret -}}
```
niv-ez, line 69: need to add a relevant explanation/comment in the values.yaml file for .existingSecret - it can be used for multiple secrets

niv-ez (approved):

PR #61, opened 2022-02-20
helm chart

d-goro:


liorhasson:
@d-goro you need to have README.md file with instructions

charts/akeyless-csi-provider/values.schema.json
```diff
@@ -0,0 +1,177 @@
+{
```
liorhasson, line 1: whats the purpose of this file?

PR #67, opened 2022-03-22
update version

renanaAkeyless:


charts/akeyless-k8s-secrets-injection/values.yaml
```diff
@@ -44,6 +44,11 @@ env:
  # AKEYLESS_API_KEY: "<api_key>"
  # AKEYLESS_CRASH_POD_ON_ERROR: "enable"
  # AKEYLESS_K8S_AUTH_CONF_NAME: "K8s_conf_name"
+
+  AKEYLESS_AGENT_LIMITS_CPU: "500m"
```
niv-ez, line 48: Why does the resources limits and requests are ENV vars instead of regular vaules?

renanaAkeyless: we do have a "regular vaules" (please at the  resources: section ) for our main webhook deployment
BUT this values are not for the webhook pod ... this values are added to all new pods that will be created (the agents)
(its MutatingWebhookConfiguration...  its job is to change the pods configuration)

niv-ez (approved):

PR #86, opened 2022-06-08
Add possibility to use existing serviceAccout

tapuhi:


charts/akeyless-api-gateway/values.yaml
```diff
@@ -19,6 +19,7 @@ deployment:
   annotations: {}
   service_account:
     enabled: false
+    use_existing: false
```
OriBenHur-akeyless, line 22: What dose use existing do?
You should state in the deployment/Statefulset the service account either way the only difference would be either to create a serviceaccount or not.

I'll open a pr to your branch to implement the change

OriBenHur-akeyless:
https://github.com/akeylesslabs/helm-charts/pull/87

charts/akeyless-api-gateway/values.yaml
tapuhi: See my comment in the pr . We don't ether way to create the service account we want to use an existing one if it already exists.

liorhasson: @tapuhi I like this setup from Vault:
https://github.com/hashicorp/vault-helm/blob/main/values.yaml#L789

I think instead of `use_existing` use `create: true` (its more clear), and also if the user choose create `true` we and leaves the name empty, we need to auto generate the SA name:
`If not set and create is true, a name is generated using the fullname template`

OriBenHur-akeyless (approved):

charts/akeyless-api-gateway/values.yaml
tapuhi: @liorhasson You mean use existing instead of both "enable" and "use existing"  and use just one "create" and only of the name is not empty we will not create one but using one?

tapuhi: @liorhasson I think it can be a bit confusing as if someone will try to add a name he will expect the helm to create a service account with this name rather than using one that does not exist yet

liorhasson: > @liorhasson You mean use existing instead of both "enable" and "use existing" and use just one "create" and only of the name is not empty we will not create one but using one? @OriMankali WDYT ?

no I meant, leave the `enable` field, and change "use existing" to "create".

liorhasson: > @liorhasson I think it can be a bit confusing as if someone will try to add a name he will expect the helm to create a service account with this name rather than using one that does not exist yet

In same cases it common to auto generate names, if the user didn't explicit add one (like in the link from vault chart)
but its not critical

tapuhi: ow ok now I get it .  will fix it .

OriBenHur-akeyless: @tapuhi just make sure to do the corresponding change to the statefulset and deployment

liorhasson (approved):

OriBenHur-akeyless (approved):

PR #87, opened 2022-06-08
Use existing service account in deployment and statefulSet

OriBenHur-akeyless:
a small patch to use the service account in statefulset and deployment


tapuhi:
Not good , as If enabled it will create the service account which we don't want as the whole purpose is to import one they already created . And in case they already created one it will fail the upgrade .

tapuhi (approved):

PR #95, opened 2022-07-08
Update values.yaml

OriBenHur-akeyless:


liorhasson (approved):

charts/akeyless-api-gateway/values.yaml
```diff
@@ -2,6 +2,9 @@
 
 replicaCount: 2
 
+# akeylessStrictMode need to be enabled when using Redhat Openshift
+akeylessStrictMode: false
```
niv-ez, line 6: Where is the template reference to this value?

OriBenHur-akeyless: this property was missing only in the values.yaml
this is the part who uses it 
https://github.com/akeylesslabs/helm-charts/blob/c266eb7e8de1ec3298f523225e2b1b58aa50ddb3/charts/akeyless-api-gateway/templates/deployment.yaml#L103

niv-ez (approved):

PR #108, opened 2022-08-15
Add allowedBastionUrls chart value

rani-sharim:
Add `allowedBastionUrls` chart value and `ALLOWED_BASTION_URLS` env var to web bastion

### DIFF
--- a/charts/akeyless-secure-remote-access/values.yaml
+++ b/charts/akeyless-secure-remote-access/values.yaml
@@ -21,6 +21,9 @@ legacySigningAlg: "false"
 # Optional, provide a key-name to extract the username (value) of that key in sub-claims. The extracted username will be used to authenticated to the remote target (RDP or SSH).
 usernameSubClaim:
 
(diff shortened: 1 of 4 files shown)


charts/akeyless-secure-remote-access/values.yaml
```diff
@@ -21,6 +21,9 @@ legacySigningAlg: "false"
 # Optional, provide a key-name to extract the username (value) of that key in sub-claims. The extracted username will be used to authenticated to the remote target (RDP or SSH).
 usernameSubClaim:
 
+# Comma separated list of the URLs that will be considered valid for redirection to this bastion (security measure to prevent Open Redirect)
+allowedBastionUrls:
+
```
niv-ez, line 26: it might better to put here a comment with an example to the expected input

rani-sharim: added example

niv-ez (approved):

liorhasson (approved):

PR #110, opened 2022-08-18
Asm 3867 remove old helm charts ssh ztb

omriezra:
Updated bash script to update ztb and ssh appVersions on sra chart, to remove old helm charts that not in use


.github/scripts/latest_version_release.sh
```diff
@@ -28,47 +26,40 @@ pushd "$GITHUB_WORKSPACE/charts/$chart_dir"
     chart_version=$(grep '^version:[[:space:]][[:digit:]]' Chart.yaml | awk '{print $2}') || die "Failed to retrieve current chart version"
     bump_version "${chart_version}" "${major_minor_patch}"
     new_chart_version=$new_version && echo "The new Chart version is: ${new_chart_version}"
-    sed -i "s/version:.*/version: ${new_chart_version}/g" Chart.yaml
+    sed  -i "s/version:.*/version: ${new_chart_version}/g" Chart.yaml
```
OriBenHur-akeyless, line 29:
```suggestion
    sed -i "s/version:.*/version: ${new_chart_version}/g" Chart.yaml
```

.github/scripts/latest_version_release.sh
```diff
@@ -28,47 +26,40 @@ pushd "$GITHUB_WORKSPACE/charts/$chart_dir"
...
 
-if [[ -n "${sra_chart}" ]]; then
-    pushd "$GITHUB_WORKSPACE/charts/$sra_chart"
-        # bump chart version
-        chart_version=$(grep '^version:[[:space:]][[:digit:]]' Chart.yaml | awk '{print $2}') || die "Failed to retrieve current chart version"
-        bump_version "${chart_version}" "${major_minor_patch}"
-        new_chart_version=$new_version && echo "The new Chart version is: ${new_chart_version}"
-        sed -i "s/version:.*/version: ${new_chart_version}/g" Chart.yaml
+      git add -A && git commit -m "Updated ${service} helm chart version to latest: ${app_version}" || die "Failed to commit changes to git"
+      git push origin HEAD
```
OriBenHur-akeyless, line 49: This approach is very fragile, you must do `git pull` before committing anything 
Moreover, the push origin will only work if the branch exist on the remote origin

omriezra: remember this run on GitHub action, so you have a checkout action before this. in addition this workflow is triggered from other repo, that is also checks out the code

.github/scripts/latest_version_release.sh
omriezra: why is that space needed?

.github/scripts/latest_version_release.sh
OriBenHur-akeyless: Sure, but what if someone marge something before the action gets around to doing its thing?

.github/scripts/latest_version_release.sh
OriBenHur-akeyless: it's not I've removed it in the suggestion

.github/scripts/latest_version_release.sh
omriezra: There is a previous step on the workflow:

`      - name: Fetch history
        run: git fetch --prune --unshallow`
        
Those this cover the issue you talk about?

.github/scripts/latest_version_release.sh
```diff
@@ -28,47 +26,40 @@ pushd "$GITHUB_WORKSPACE/charts/$chart_dir"
...
-        # edit app version
-        ztb_app_ver=$(grep appVersion ../akeyless-zero-trust-bastion/Chart.yaml | awk '{print $2}')
-        ssh_app_ver=$(grep appVersion ../akeyless-ssh-bastion/Chart.yaml | awk '{print $2}')
-        sed -i "s/appVersion.*/appVersion: ${ztb_app_ver}_${ssh_app_ver}/g" Chart.yaml
-        
+    else
+        sed -i "s/appVersion.*/appVersion: ${app_version}/g" Chart.yaml
+    fi
+
+    if [[ "${chart_dir}" != "akeyless-secure-remote-access"  ]]; then
```
niv-ez, line 58: Why? you can use the else statement on line 54 to do the git add and push since you already know its not "akeyless-secure-remote-access" dir

.github/scripts/latest_version_release.sh
omriezra: updated

.github/scripts/latest_version_release.sh
omriezra: updated

.github/scripts/latest_version_release.sh
omriezra: updated

niv-ez (approved):

OriBenHur-akeyless (approved):

PR #111, opened 2022-08-28
Add FIPS support to api-gateway helm chart

OriBenHur-akeyless:


charts/akeyless-api-gateway/values.yaml
```diff
@@ -36,6 +36,7 @@ deployment:
   #      eks.amazonaws.com/role-arn: arn:aws:iam::40XXXXXXXX75:role/eksctl-Role1-XXXXXXXX
   nodeSelector:
     #     iam.gke.io/gke-metadata-server-enabled: "true"
+  fips: false
```
niv-ez, line 39: I think it is better to use the following, since its more informative 
fips:
  enabled: false

also add a comment with description before like:
# enable FIPS mode to.....

OriBenHur-akeyless: @niv-ez Done

niv-ez (approved):

liorhasson (approved):

PR #123, opened 2022-10-16
fix Service Account settings location on sra chart

omriezra:


OriBenHur-akeyless (approved):

charts/akeyless-secure-remote-access/values.yaml
```diff
@@ -42,6 +42,15 @@ privilegedAccess:
   allowedAccessIDs: []
    # - p-1234
    # - p-5678
+  service_account:
```
niv-ez, line 45: use "serviceAccoun" instead, it's Helm values convention to use camel case and one-word attributes without "_" lowercases

charts/akeyless-secure-remote-access/templates/_helpers.tpl
```diff
@@ -104,11 +104,11 @@ Checks persistent volume
 Get serviceAccountName
 */}}
 {{- define "akeyless-api-gw.getServiceAccountName" -}}
-    {{- if and (not .Values.sshConfig.service_account.serviceAccountName) ( not .Values.sshConfig.service_account.create )  }}
```
liorhasson, line 107: please verify with @OriMankali if we need to support **backward computability** here

omriezra: resolved

OriBenHur-akeyless (approved):

niv-ez (approved):

PR #156, opened 2023-01-18
Fixing customerFragmentExist value in helpers.tpl

idan-gur:
Now checks also for the customer fragment encoded value that was included in release 1.19.3


idan-gur:
@liorhasson 
Forgot to update it in the helpers so the new value was not defined inside the deployment.yaml files.
We checked it locally and now all is working properly.

Thanks in advance!

liorhasson (approved):

niv-ez (approved):

idan-gur:
@liorhasson Can you please merge the PR?
Thank you!

PR #166, opened 2023-03-03
Support to modify MutatingWebhookConfiguration failurePolicy and namespace/object selectors

alvarof2:
Adds the ability to modify `failurePolicy` and to add namespace/object selector `matchExpressions` to the `MutatingWebhookConfiguration` object of the `k8s-secrets-injection` Helm chart; while respecting the default values.


jcortejoso:
@liorhasson Can you take a look?

PR #177, opened 2023-05-01
Fix non-root cp error

OriBenHur-akeyless:
The chart was using hardcoded `/root` we replaced it with a function to determine config root path to avoid access denied error

### DIFF
--- a/charts/akeyless-api-gateway/templates/_helpers.tpl
+++ b/charts/akeyless-api-gateway/templates/_helpers.tpl
@@ -111,6 +111,14 @@ Generate chart secret name
 {{/*
 Check customer fragment
 */}}
+
+{{- define "akeyless-api-gw.root.config.path" -}}
+{{- if .Values.akeylessStrictMode }}
+     {{- printf "/home/akeyless" -}}
+{{- else }}
+     {{- printf "/root" -}}
+{{- end -}}
+{{- end -}}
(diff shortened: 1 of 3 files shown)


niv-ez (approved):

charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -111,6 +111,14 @@ Generate chart secret name
 {{/*
 Check customer fragment
 */}}
+
+{{- define "akeyless-api-gw.root.config.path" -}}
+{{- if .Values.akeylessStrictMode }}
+     {{- printf "/home/akeyless" -}}
```
liorhasson, line 117: @renanaAkeyless please approve this path

OriBenHur-akeyless: I've talked with @barucoh and we saw that the code first chack the user home dir and if it can't find the files there it will check `/root`

barucoh: If the username is always `akeyless` than this should be fine, but I would use:
```
printf "$HOME"
```
I think this should always work btw, also for `root`...

barucoh (approved):
One comment, but if `akeyless` is always the user than approved

PR #184, opened 2023-06-11
Add flexibility for TLS certificate and key file names

devorbitus:
This PR aims to provide more flexibility to the users of the akeyless-api-gateway Helm chart by allowing them to specify the names of the TLS certificate and key files.

Details:

Previously, the names of the certificate and key files were hardcoded as akeyless-api-cert.crt and akeyless-api-cert.key respectively. This became a problem when users wanted to use certificates generated by cert-manager, which uses the names tls.crt and tls.key for the certificate and key files respectively.

To solve this problem, we introduced two new properties in the values.yaml file.

.Values.TLSConf.certFileName
.Values.TLSConf. keyFileName

The required function is used to ensure that the deployment won't proceed unless these values are provided. If either TLSConf.certFileName or TLSConf.keyFileName are not provided, the Helm deployment will fail with an error message. This change ensures that the chart remains robust while providing additional flexibility to the users.

This change is backward compatible. If no file names are provided, the chart will default to using akeyless-api-cert.crt and akeyless-api-cert.key, maintaining the previous behavior.

This change will greatly improve the usability of the chart with automated certificate managers like cert-manager and make the chart more flexible for various user setups.

### DIFF
--- a/charts/akeyless-api-gateway/templates/deployment.yaml
+++ b/charts/akeyless-api-gateway/templates/deployment.yaml
@@ -98,10 +98,10 @@ spec:
               cp /logand_conf/logand.conf {{include "akeyless-api-gw.root.config.path" $}}/.akeyless/logand.conf
             {{- end }} 
             {{- if eq (include "akeyless-api-gw.tlsCertificateExist" .) "true"}}
-              cp /tls_conf/{{ required "TLSConf.certFileName is required" .Values.TLSConf.certFileName }} {{include "akeyless-api-gw.root.config.path" $}}/.akeyless/akeyless-api-cert.crt
+              cp /tls_conf/{{ include "akeyless-api-gw.certFileName" . }} {{include "akeyless-api-gw.root.config.path" $}}/.akeyless/akeyless-api-cert.crt
             {{- end }} 
             {{- if eq (include "akeyless-api-gw.tlsPrivateKeyExist" .) "true"}}
-              cp /tls_conf/{{ required "TLSConf.keyFileName is required" .Values.TLSConf.keyFileName }} {{include "akeyless-api-gw.root.config.path" $}}/.akeyless/akeyless-api-cert.key
+              cp /tls_conf/{{ include "akeyless-api-gw.keyFileName" . }} {{include "akeyless-api-gw.root.config.path" $}}/.akeyless/akeyless-api-cert.key
             {{- end }} 
           volumeMounts:
           {{- if eq (include "akeyless-api-gw.customerFragmentExist" .) "true"}}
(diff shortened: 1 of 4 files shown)


devorbitus:
This change is primarily for Walmart

charts/akeyless-api-gateway/templates/deployment.yaml
OriBenHur-akeyless, line 1: This was done with appropriate changes to the backend?
I think they are looking for this file names.
@renanaAkeyless ?

devorbitus: no changes to the backend are required because of the init container that we use to copy the input file (read-only) to the writable location to be used by the gateway

devorbitus: This PR will alter the INPUT file name from a static value to a configurable one

liorhasson: I'm not sure it will work, the backend looks for the file name: `akeyless-api-cert.crt` hardcoded in the base code

devorbitus: > I'm not sure it will work, the backend looks for the file name: `akeyless-api-cert.crt` hardcoded in the base code

I agree that the backend looks for the files to be named specific names. 

Why are we using an init container? We use an init container to copy from a read only volume mount into a writable volume mount. It's not changing the writable side that the gateway uses AFTER the init container copies the files, it's only changing the read only file BEFORE the init container copies the file.

OriBenHur-akeyless: You are right my bad, as long the target file stays the same the source file can be anything.
I would just suggest having a default value

OriBenHur-akeyless (approved):

charts/akeyless-api-gateway/templates/deployment.yaml
devorbitus: The values file has the default value per helm website

charts/akeyless-api-gateway/values.yaml
```diff
@@ -294,10 +294,14 @@ TLSConf:
   enableSniProxy: false
   # minimumTlsVersion can be one of the following <TLSv1/TLSv1.1/TLSv1.2/TLSv1.3> 
   minimumTlsVersion:
-  # Specifies an existing secret for tls-certificate must include:
-  # - akeyless-api-cert.crt (base64)
```
niv-ez, line 298: tlsExistingSecretName is looking for an existing k8s secret with these field and structure - otherwise, it will not work. so it need to be mentioned , also you need to check for backwards compatibility to support both old and new deployment ways without breaking or changing his values file

devorbitus: Hey @niv-ez these lines copy the input file into the output file that is actually used by the gateway https://github.com/akeylesslabs/helm-charts/blob/existing-secret-input-file-override/charts/akeyless-api-gateway/templates/deployment.yaml#L100-L105

OriBenHur-akeyless (approved):

liorhasson (approved):

PR #189, opened 2023-07-02
Added support for autoscaling/v2 in kube 1.26

OriBenHur-akeyless:


charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -365,4 +365,12 @@ Check metrics configuration Secret
     {{- else -}}
         {{- printf "false" -}}
     {{- end -}}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{- define "hpa.api.version" }}
+    {{- if .Capabilities.APIVersions.Has "autoscaling/v2beta2" }}
```
niv-ez, line 371: I think that the validation should check first if the new API ver exists than use it ("autoscaling/v2") if not, check the older beta one

OriBenHur-akeyless: this will make all the apps to go out of sync

niv-ez (approved):

PR #198, opened 2023-08-01
added security context on deployment level and changed security conte…

michelsk:
…xt position to include default values that are needed and the option to add more values using --set

### DIFF
--- a/charts/akeyless-k8s-secrets-injection/values.yaml
+++ b/charts/akeyless-k8s-secrets-injection/values.yaml
@@ -14,11 +14,17 @@ gatewayCert:
   # - tls.crt (base64)
   tlsCertsSecretName:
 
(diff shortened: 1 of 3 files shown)


charts/akeyless-k8s-secrets-injection/values.yaml
```diff
@@ -4,6 +4,10 @@
 
 deployment:
   labels: {}
+  # Remove the {} and add any needed values to your SecurityContext
```
renanaAkeyless, line 7: why do i see SecurityContext twice?

michelsk: one is on the deployment level and the other is on the containers level

liorhasson (approved):

PR #205, opened 2023-09-12
hpa

renanaAkeyless:


charts/akeyless-k8s-secrets-injection/templates/_helpers.tpl
```diff
@@ -41,3 +41,12 @@ Create the name of the service account to use
...
 {{- end -}}
+
+
+{{- define "hpa.api.version" }}
+    {{- if .Capabilities.APIVersions.Has "autoscaling/v2" }}
+        {{- printf "autoscaling/v2" }}
+    {{- else }}
+        {{- printf "autoscaling/v2beta2" }}
+    {{- end }}
+{{- end }}
```
OriBenHur-akeyless, line 52: We need to make sure this approach works in argocd, something is telling me it won't work

OriBenHur-akeyless (approved):

charts/akeyless-k8s-secrets-injection/templates/_helpers.tpl
renanaAkeyless: why should i care for argocd? this helm is for our customers

niv-ez: it's ok just BC for HPA version, worst case it will take a default one if it's not deprecated

niv-ez (approved):

charts/akeyless-k8s-secrets-injection/templates/_helpers.tpl
OriBenHur-akeyless: You should care about argocd because we have customers who use it.
About the default, it's deprecated since kube version 1.23

PR #211, opened 2023-10-03
Allow users to exclude cipher suites

avishay28:
Add env variable to allow users to provide list of tls cipher suites to exclude


niv-ez:
Add the change to the Readme

avishay28:
@niv-ez we don't have any section in the readme that talk about the TLS parameters, should i add new section just for this param?

avishay28:
@niv-ez added a new section for general config 
https://github.com/akeylesslabs/helm-charts/pull/211/commits/bd577beebc0e60fbfe371042794c8e60b4ffcf9d

niv-ez (approved):

renanaAkeyless (approved):

PR #216, opened 2023-11-09
gw and injector as DaemonSet ASM-6249

renanaAkeyless:
https://akeyless.atlassian.net/browse/ASM-6249

### DIFF
--- a/charts/akeyless-api-gateway/templates/deployment.yaml
+++ b/charts/akeyless-api-gateway/templates/deployment.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: {{ .Values.kind }}
 metadata:
   name: {{ include "akeyless-api-gw.fullname" . }}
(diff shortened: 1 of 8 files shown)


charts/akeyless-api-gateway/templates/deployment.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: {{ .Values.kind }}
```
niv-ez, line 2: why? this template is "deployment" kind by definition

renanaAkeyless: this customer wants to run the gw/injector as DeamonSet and not Deployment
https://akeyless.atlassian.net/browse/ASM-6249

niv-ez: OK, so it is better to create a flag on values.yaml file and make a different yaml for Daemonset so it will be easier to maintain.

PR #224, opened 2023-11-26
Update ZTP configuration

michelsk:


liorhasson (approved):

charts/akeyless-secure-remote-access/templates/akeyless-zero-trust-portal/configmap.yaml
```diff
@@ -10,6 +10,9 @@ metadata:
 data:
   env-config.js: |
     window._env_ = {
-        GW_DNS: {{ $.Values.apiGatewayURL | quote }}
+        CLI_GW_DNS: {{ $.Values.apiGatewayURL | quote }},
+        AUTH_DNS: 'https://auth.akeyless.io',
```
niv-ez, line 14: You can put AUTH_DNS and the rest of the vars as default values with an option to change so other tenant could be applied

michelsk: @liorhasson should I leave it hardcoded as is or move it to values file with default ?

liorhasson: Leave it, FE will add the option to work with other tenants

liorhasson (approved):

niv-ez (approved):

PR #229, opened 2023-12-28
restat-rollout

renanaAkeyless:


charts/akeyless-k8s-secrets-injection/templates/webhook-servicaccount.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.serviceAccount.create -}}
+{{- if or (.Values.serviceAccount.create) (.Values.restartRollout.enabled) }}
```
niv-ez, line 1: what if the user entered his own SA (and used the "name" field)? will the new role binding match the SA?

niv-ez (approved):

omriezra (approved):

PR #233, opened 2024-02-21
Added support for DaemonSet for akeyless gateway

OriBenHur-akeyless:
This pr aims to add the ability to deploy Akeyless Gateway as a DaemonSet instead of as a deployment

### DIFF
--- a/charts/akeyless-api-gateway/templates/deployment.yaml
+++ b/charts/akeyless-api-gateway/templates/deployment.yaml
@@ -1,5 +1,11 @@
 apiVersion: apps/v1
+{{- if eq .Values.kind "Deployment" }}
 kind: Deployment
+{{- else if eq .Values.kind "DaemonSet" }}
+kind: DaemonSet
+{{- else }}
+kind: Deployment
+{{- end }}
 metadata:
   name: {{ include "akeyless-api-gw.fullname" . }}
   namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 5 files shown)


ak-nir (approved):

charts/akeyless-api-gateway/templates/deployment.yaml
```diff
@@ -1,5 +1,11 @@
 apiVersion: apps/v1
+{{- if eq .Values.kind "Deployment" }}
```
niv-ez, line 2: Deployment should be default value and change to DaemonSet only if the user asked for it

charts/akeyless-api-gateway/templates/deployment.yaml
```diff
@@ -13,8 +19,7 @@ metadata:
   {{- toYaml .Values.deployment.annotations | nindent 4 }}
   {{- end }}
 spec:
-
-  {{- if not .Values.HPA.enabled }}
+  {{- if and (eq .Values.kind "Deployment") (not .Values.HPA.enabled) }}
```
niv-ez, line 22: what with HPA yaml?
If set to DaemonSet the HPA should not be applied on YAML level as well

charts/akeyless-api-gateway/templates/deployment.yaml
OriBenHur-akeyless: It's the default see the else at the end of the condition

charts/akeyless-api-gateway/templates/deployment.yaml
OriBenHur-akeyless: You are right i'll fix it

niv-ez (approved):

PR #252, opened 2024-08-11
Feature/asm 11750 redis tls

dan-akeyless:
- [ ] semver needs to be checked for gw versions that support redis tls in `cache_enable_tls`


charts/akeyless-api-gateway/values.yaml
```diff
@@ -18,6 +18,9 @@ initContainer:
     pullPolicy: IfNotPresent
 
 cache:
+  tls:
+    autoGenerated: false
+    secretName: ""
```
renanaAkeyless, line 23: remove secretName - for autoGenerated please generate a name...
please add existingSecretName

renanaAkeyless (changes requested):
remove secretName - for autoGenerated please generate a name...
please add existingSecretName

charts/akeyless-api-gateway/values.yaml
dan-akeyless: secretName is existingSecretName, will rename

charts/akeyless-api-gateway/templates/deployment.yaml
```diff
@@ -172,6 +184,8 @@ spec:
           resources:
 {{- toYaml .Values.resources | nindent 12 }}
           env:
+            # - name: DEVELOPMENT_OR_RELEASE
+            #   value: "development"
```
eitan-kr, line 188: We shouldn't expose this to the customers

charts/akeyless-api-gateway/values.yaml
```diff
@@ -396,9 +399,11 @@ cachingConf:
 
   clusterCache:
     enabled: false
+    cacheAddress: ""
```
renanaAkeyless, line 402: remove cacheAddress - the client should not be exposed to it.. its internal service

charts/akeyless-api-gateway/values.yaml
```diff
@@ -36,7 +39,7 @@ image:
 ##   Alternative mirror registry
 #  repository: docker.registry-2.akeyless.io/base
 
-  pullPolicy: Always
+  pullPolicy: Never
```
renanaAkeyless, line 42: remove Never

charts/akeyless-api-gateway/values.yaml
```diff
@@ -18,9 +18,12 @@ initContainer:
     pullPolicy: IfNotPresent
 
 cache:
+  tls:
+    autoGenerated: false
+    existingSecretName: ""
   image:
     repository: docker.io/bitnami/redis
-    tag: 6.2
+    tag: 7.4.0
```
renanaAkeyless, line 26: why did you switched to 7.4.0?

charts/akeyless-api-gateway/values.yaml
```diff
@@ -18,9 +18,12 @@ initContainer:
     pullPolicy: IfNotPresent
 
 cache:
+  tls:
+    autoGenerated: false
+    existingSecretName: ""
```
renanaAkeyless, line 23: duplicate with existingCacheConnectionSecret??

charts/akeyless-api-gateway/values.yaml
dan-akeyless: the customer wants to connect to his aws elasticache instance

renanaAkeyless: product did not approve this... who told you to do it? this is not a valid request

charts/akeyless-api-gateway/values.yaml
dan-akeyless: reverted

charts/akeyless-api-gateway/values.yaml
dan-akeyless: mistake, reverted

charts/akeyless-api-gateway/values.yaml
dan-akeyless: one is for tls the other is the redis password

charts/akeyless-api-gateway/values.yaml
dan-akeyless: eldad?

charts/akeyless-api-gateway/templates/cache.yaml
```diff
@@ -60,40 +72,65 @@ spec:
...
+          value: '--maxmemory {{ include "akeyless-api-gw.redisMaxmemory" . }} --maxmemory-policy allkeys-lru'
         - name: REDIS_PASSWORD
           valueFrom:
             secretKeyRef:
               name: {{ include "akeyless-api-gw.cacheSecretName" . }}
               key: cache-pass
+        - name: REDIS_TLS_ENABLED
+          value: {{ include "cache_enable_tls" . | quote -}}
+        {{- if (include "cache_enable_tls" . | eq "true") }}
+        - name: REDIS_TLS_PORT
```
renanaAkeyless, line 99: why do we need REDIS_TLS_PORT and REDIS_PORT ? shouldn't we have only one port tls or not?

dan-akeyless: Both should be configured in the case TLS is enabled. 
The svc only exposes the TLS port

charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -416,17 +416,33 @@ Check metrics configuration Secret
...
 /health
 {{- else -}}
 /
 {{- end }}
 {{- end }}
-{{- end }}
\ No newline at end of file
+
+{{- define "cache_enable_tls" -}}
+{{ and .Values.cache .Values.cache.tls (or .Values.cache.tls.autoGenerated (ne .Values.cache.tls.existingSecretName "")) }}
```
renanaAkeyless, line 427: this condition is not accurate we need to look at clusterCache.enabled and .Values.cache.tls is not a value ?
please fix

charts/akeyless-api-gateway/templates/tls.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if ( and .Values.cache .Values.cache.tls .Values.cache.tls.autoGenerated) }}
+{{- $fullname := .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- $ca := genCA "redis-ca" 365 }}
+{{- $altNames := list $fullname "127.0.0.1" "localhost" (include "cache_address" .) (printf "*.%s" (include "cache_address" . ))}}
```
renanaAkeyless, line 4: remove localhost... we dont support external redis only the one we create here

dan-akeyless: Localhost is not relevant to external redis. This certificate is only relevant for use in this cluster only

charts/akeyless-api-gateway/templates/_helpers.tpl
dan-akeyless: .Values.cache.tls is indeed not a value but I still need to traverse it otherwise I might get a nil pointer reference on .Values.cache.tls.autoGenerated 

clusterCache.enabled is not relevant here since the configuration of the redis cache is encapsulated:
```
            {{- if or (eq (include "akeyless-api-gw.adminAccessUidExist" .) "true") ( .Values.cachingConf.clusterCache.enabled) }}
            - name:  GATEWAY_CLUSTER_CACHE
              value: enable
            - name: ENABLE_CACHE_TLS
              value: {{ include "cache_enable_tls" . | quote }}
```

omriezra: i have no problem with the logic/condition but the ENABLE_CACHE_TLS and it's config should be under cluster cache configuration to prevent user mistakes, with that said i think product should decide where it should be located @renanaAkeyless

charts/akeyless-api-gateway/templates/tls.yaml
omriezra, line 1: i would change to redis-tls.yaml

charts/akeyless-api-gateway/templates/tls.yaml
```diff
@@ -0,0 +1,23 @@
...
+{{- $cert := genSignedCert $fullname nil $altNames 365 $ca }}
+{{- $existingSecret := (lookup "v1" "Secret" .Release.Namespace (include "generated_cache_tls_secret_name" .)).data }}
```
omriezra, line 6: i suggest to to remove lookup or to mention in the readme / docs that if client using gitops tools, they can only used with existing secret. please sync with product @renanaAkeyless

charts/akeyless-api-gateway/templates/cache.yaml
```diff
@@ -60,40 +72,65 @@ spec:
...
+          value: {{ $tlsCACertPath }}
+        {{ else }}
+        - name: REDIS_PORT
+          value: "6379"
+        {{- end }}
+        {{- if (include "cache_enable_tls" . | eq "true") }}
+        volumeMounts:
+            - name: cache-tls
+              mountPath: {{ $mountCertPaths }}
+              readOnly: true
```
omriezra, line 119: verify indentation

charts/akeyless-api-gateway/templates/tls.yaml
dan-akeyless: done

charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -416,17 +416,33 @@ Check metrics configuration Secret
 {{- end }}
 
 {{- define "health_check_path" -}}
```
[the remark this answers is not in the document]
dan-akeyless: done

charts/akeyless-api-gateway/templates/cache.yaml
dan-akeyless: done

charts/akeyless-api-gateway/templates/tls.yaml
dan-akeyless: Whether the lookup is there or not it will still reconcile and recreate the files, so I recommend we document it
@EldadH89 how do you want to formulate it in the readme?

charts/akeyless-api-gateway/templates/tls.yaml
renanaAkeyless: we don't allow external redis! please remove

charts/akeyless-api-gateway/values.yaml
```diff
@@ -18,6 +18,9 @@ initContainer:
...
```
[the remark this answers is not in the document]
renanaAkeyless: for the new chart.. in the old one we don't change it now

charts/akeyless-api-gateway/templates/tls.yaml
dan-akeyless: Again, this has no connection to external redis. This certificate is only relevant for use in this cluster only

dan-akeyless: Removed "127.0.0.1" "localhost" 
The customer can still access the cache if they exec into the redis pod even though "127.0.0.1" "localhost" were removed as per my tests

omriezra (approved):

renanaAkeyless (approved):

PR #265, opened 2024-10-29
Support new approach for storing gateway cache encryption key in k8s secrets

idanmantin:
## Description

Task: [ASM-12592 - Redis Enc Key Management In SaaS Gator](https://akeyless.atlassian.net/browse/ASM-12592)

## Changes:
- New field: `storeEncryptionKeyToK8sSecrets` - When set to `true`, the cache encryption key is stored in K8s secrets, supporting Auto Scaling scenarios without network connectivity.

- Enhancements to field `encryptionKeyExistingSecret`:
  - Still specifies the K8s secret name to be used for managing the cache encryption key.
  - If empty, the chart will generate name based on the deployment name
  - **The Gateway will use and manage this secret’s content**, will create/update the k8s secret with the cache enc key, useful for Scaling Up without network connectivity.
- When chart creating the **Service Account**, we also take care to create a **Role** and **Role Binding** to the created Service Account with appropriate **K8s secret permissions** (get, create, update) on this specific secret name, allowing the gateway to manage the cache encryption key securely in Kubernetes.

### DIFF
--- a/charts/akeyless-api-gateway/templates/_helpers.tpl
+++ b/charts/akeyless-api-gateway/templates/_helpers.tpl
@@ -122,10 +122,8 @@ Generate chart secret name
 {{- define "akeyless-api-gw.clusterCacheEncryptionKeyExist" -}}
     {{- if .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
         {{- printf "%s" .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
-    {{- else -}}
-        {{- if .Values.cachingConf.clusterCache.enabled }}
-            {{- printf "%s-cache-encryption-key" .Release.Name -}}
-        {{- end -}}
+    {{- else if .Values.cachingConf.clusterCache.enabled -}}
+        {{- printf "%s-cache-encryption-key" .Release.Name -}}
     {{- end -}}
 {{- end -}}
 
--- a/charts/akeyless-api-gateway/templates/_helpers.tpl
+++ b/charts/akeyless-api-gateway/templates/_helpers.tpl
@@ -120,9 +120,13 @@ Generate chart secret name
 {{- end -}}
 
 {{- define "akeyless-api-gw.clusterCacheEncryptionKeyExist" -}}
-        {{- if .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
-            {{- printf "%s" .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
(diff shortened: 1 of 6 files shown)


peter-akeyless (approved):

charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -120,9 +120,13 @@ Generate chart secret name
...
 {{- define "akeyless-api-gw.clusterCacheEncryptionKeyExist" -}}
-        {{- if .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
-            {{- printf "%s" .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
+    {{- if .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
+        {{- printf "%s" .Values.cachingConf.clusterCache.encryptionKeyExistingSecret -}}
+    {{- else -}}
+        {{- if .Values.cachingConf.clusterCache.enabled }}
+            {{- printf "%s-cache-encryption-key" .Release.Name -}}
         {{- end -}}
+    {{- end -}}
```
omriezra, line 129:
```suggestion
    {{- else if .Values.cachingConf.clusterCache.enabled }}
      {{- printf "%s-cache-encryption-key" .Release.Name -}}
    {{- end -}}
```

omriezra (dismissed):

charts/akeyless-api-gateway/templates/_helpers.tpl
idanmantin: Fixed.

charts/akeyless-api-gateway/templates/serviceAccount.yaml
```diff
@@ -8,6 +9,42 @@ metadata:
...
   annotations:
     {{- toYaml .annotations | nindent 4 }}
-  {{- end -}}
-{{- end -}}
-{{- end -}}
\ No newline at end of file
+  {{- end }}
+
+  {{/* RBAC rule to access the secret if storeEncryptionKeyToK8sSecrets is true and clusterCacheEncryptionKeyExist is not empty */}}
+  {{- if and (eq $.Values.cachingConf.clusterCache.storeEncryptionKeyToK8sSecrets true) (not (eq (include "akeyless-api-gw.clusterCacheEncryptionKeyExist" $) "")) }}
```
omriezra, line 15: this condition needs to be update to support all the cases:

- if secret exist
- if secret not exist

should basically be create if cache enabled
@rani-sharim

OriBenHur-akeyless (approved):

OriBenHur-akeyless (approved):

PR #272, opened 2024-12-15
[ASM-13081] k8s injector - creds issue

meravit:
on every helm update, new tls cert is created for injector. It is used for communication with k8s API (mutatingwebhookconfiguration webhook).

problem:
on helm update, "old" pods are getting out of sync while they are still serving request till the new pod will be ready.
because new cert is created, we update webhook with new ca, while old pod are working with the old certs.

solution:
pass secret name to injector so it will listen to the change and update tls on the fly

**changes:**
**1. rename secret keys to standard names**
**2. pass secret name and namespace to injector**
**3. add permission for injector to read k8s secret**

see injector PR
https://github.com/akeylesslabs/akeyless-kubernetes-webhook-secrets/pull/74


peter-akeyless (approved):

meravit:
@akeylesslabs/devops please review

peter-akeyless (approved):

ofirattiasakeyless (approved):

PR #275, opened 2024-12-30
Added EXTERNAL_SSH_PORT environment in ssh bastion for using in /status rest call

alikdolg:


avishay28 (approved):

meravit (approved):

idohamburger (approved):

charts/akeyless-gateway/templates/akeyless-secure-remote-access/ssh_deployment.yaml
```diff
@@ -77,6 +77,8 @@ spec:
           - containerPort: 9900
             name: curl-proxy
           env:
+            - name: EXTERNAL_SSH_PORT
```
omriezra, line 80: why this is not default to some port?

alikdolg: we always have "Values.sra.sshConfig.service.port" defined. This is no option to run SRA without ssh port

alikdolg: Added also to "old" SRA helm
```
 ports:
          - containerPort: 22
            name: ssh
          - containerPort: 9900
            name: curl-proxy
          env:
          - name: EXTERNAL_SSH_PORT
            value: "2222"
          - name: VERSION
            value: 2.3.6
          - name: CLUSTER_NAME
            value: defaultCluster
          - name: LEGACY_SIGNING_ALG
            value: "false"
          - name: BASTION_TYPE
            value: "ssh-proxy" 
```

omriezra (approved):

PR #288, opened 2025-02-17
Fix azure object ID env name ("unified" chart)

avishay28:
The bastions expect env name `AZURE_OBJECT_ID` and not `AZURE_OBJ_ID`


avishay28:
GW code expect `AZURE_OBJ_ID` and it share the same deployment with bastions so this solution is not good

PR #293, opened 2025-03-06
Support Redis password configuration + new Redis version 7.4.2

alikdolg:


charts/akeyless-api-gateway/templates/_helpers.tpl
```diff
@@ -131,7 +131,11 @@ Generate chart secret name
 {{- end -}}
 
 {{- define "akeyless-api-gw.cacheSecretName" -}}
-        {{ $.Release.Name }}-cache-secret
+    {{- if empty .Values.cachingConf.clusterCache.cachePasswordExistingSecret }}
+        {{- printf "%s-cache-secret" $.Release.Name -}}
+    {{- else if not (empty .Values.cachingConf.clusterCache.cachePasswordExistingSecret) }}
```
eitan-kr, line 136: Whet is the point of this if? you already checked this in the above if and if you got to the else, it is not empty.

idohamburger (approved):

charts/akeyless-api-gateway/templates/_helpers.tpl
eitan-kr: @omriezra ?

eitan-kr (approved):

PR #297, opened 2025-03-19
rm priviledged

avishay28:
SSH bastion doesn't need to run with both `priviledged` and `SYS_ADMIN` capabilities,  SYS_ADMIN is enough


omriezra (approved):

avishay28:
libi's tests are falling without this flag, appeearntly it's needed
Screenshot 2025-03-23 at 14 46 17

PR #310, opened 2025-06-12
update Capital and syntax

idohamburger:


OriBenHur-akeyless (changes requested):
You are missing the deployment.yaml changes

PR #311, opened 2025-06-15
Fix customCA Mount Path

OriBenHur-akeyless:


charts/akeyless-gateway/templates/deployment.yaml
ofirattiasakeyless, line 1: Don't you need to change it also in line 54 and 113?

OriBenHur-akeyless: No, the only incorrect place is the volume mount in the container
It was mounted in the wrong folder. This PR fixes that

PR #327, opened 2025-08-26
Update Bitnami Helm Repository References to Legacy

dordavid:
This PR updates the Bitnami Helm chart repository references to use the legacy repo.
The change ensures compatibility with existing deployment images.


charts/akeyless-gateway/values.yaml
```diff
@@ -135,6 +135,34 @@ cacheHA:
   ## Required for cacheHA, a storage class with the ReadWriteOnce access mode
   global:
     defaultStorageClass: ""
+    security:
+      allowInsecureImages: true  # Required for custom image repositories
+  
+  metrics:
```
dors-akeyless, line 141: if disabled by default, we can skip this one

PR #332, opened 2025-10-05
ASM-14210 expose unified GW variable sshHost

Hanan-Akeyless:
Add sra.sshHost to configure dedicated SRA SSH host

Introduces a new Helm value, `sra.sshHost`, in the `akeyless-gateway` chart to support Unified Gateway deployments where the SRA SSH service is exposed on a separate endpoint from the main Gateway API.

In environments with separate LoadBalancers for the Gateway and SSH bastion, clients need a reliable way to discover the correct SSH host. This change facilitates that by:

1.  Adding `sra.sshHost` to `values.yaml`, allowing administrators to specify the publicly accessible SSH bastion host and port (e.g., `ssh-gw.example.com:22`).
2.  Conditionally injecting this value as the `GW_SRA_SSH_HOST` environment variable into the Akeyless Gateway container.

This allows the Gateway's new host discovery endpoint to provide the correct, dedicated SSH host to clients like the Akeyless CLI, ensuring that `akeyless connect` targets the right service.

If `sra.sshHost` is not set, the system maintains its default behavior, where the Gateway host is assumed for SSH connections.

### DIFF
--- a/charts/akeyless-gateway/Chart.yaml
+++ b/charts/akeyless-gateway/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: akeyless-gateway
-version: 1.9.10
+version: 1.9.11
 description: A Helm chart for Kubernetes that deploys akeyless-gateway
 type: application
 keywords:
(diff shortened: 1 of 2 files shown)


dors-akeyless (approved):

Hanan-Akeyless:
Closed due to requirements change
This value will be injected via SaaS manage gateway page instead of yaml files

PR #372, opened 2026-04-23
chore(ztwa): added specific uid and gid to validator service

lidorb-akeyless:


charts/akeyless-zero-trust-web-access/templates/_helpers.tpl
```diff
@@ -103,4 +103,55 @@ Checks kubernetes API version support for ingress BC
...
+    {{- $names = append $names "akeyless-docker-hub-web-access" -}}
+  {{- end -}}
+{{- end -}}
+{{- if $names -}}
+imagePullSecrets:
+{{- range $names }}
+  - name: {{ . | quote }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
Alex-Souslik, line 157: Why do we need this function?

charts/akeyless-zero-trust-web-access/values.schema.json
```diff
@@ -0,0 +1,111 @@
+{
```
Alex-Souslik, line 1: you should make sure the schema allows additional fields

charts/akeyless-zero-trust-web-access/templates/_helpers.tpl
Hanan-Akeyless: not sure what this is doing, we just need imagePullSecret with the relevant data to pull the image

charts/akeyless-zero-trust-web-access/templates/NOTES.txt
Hanan-Akeyless, line 1: that's cool, how does a user see this note?

charts/akeyless-zero-trust-web-access/templates/secrets.yaml
```diff
@@ -1,11 +1,13 @@
...
-      "helm.sh/hook": pre-install
-    namespace: {{ .Release.Namespace }}
+  name: akeyless-docker-hub-web-access
+  annotations:
+    "helm.sh/hook-delete-policy": before-hook-creation
+    "helm.sh/hook": pre-install
+  namespace: {{ .Release.Namespace }}
 data:
-  .dockerconfigjson: {{ required "A valid .Values.image.dockerconfigjson entry required!" .Values.image.dockerRepositoryCreds }}
+  .dockerconfigjson: {{ .Values.image.dockerRepositoryCreds }}
```
Hanan-Akeyless, line 11: So we are creating the secret for the user based on the values.yaml? 
or they need to have it set manually and we are just using it?

charts/akeyless-zero-trust-web-access/templates/_helpers.tpl
lidorb-akeyless: the old version of this chart required a base64 string that encodes docker repo credentials.
I wanted to keep it supported for backward compatibility while allowing to supply a list of existing secrets.
Cisco also requested to be able to set a secret per continer.
So I gave all the above option and this function is supposed to be a helper that merges all of them if present

charts/akeyless-zero-trust-web-access/templates/NOTES.txt
lidorb-akeyless: if is printed in the terminal after successful install/upgrade

charts/akeyless-zero-trust-web-access/templates/secrets.yaml
lidorb-akeyless: the old behaviour where we create the secret for them is kept for backward compatibility.
but if they don't and give us a pre-existing secret we will use it

charts/akeyless-zero-trust-web-access/values.schema.json
lidorb-akeyless: I understand `"additionalProperties": true` is the default but I've added explicitly to be safe

Hanan-Akeyless (approved):

Alex-Souslik (approved):

PR #373, opened 2026-04-29
Feat/ztwa recording helm env

Hanan-Akeyless:
## Summary

Adds first-class Helm support for ZTWA session recording (worker capture +
dispatcher S3 upload) and fixes a precedence bug where the legacy
`dispatcher.config.recording` block was silently swallowing the unified
`sessionRecording.upload.*` values.

The chart now exposes a single user-facing section — `sessionRecording.*`
— that drives both the worker recorder and the dispatcher uploader, with
optional per-field overrides on either service for advanced users.

### `values.yaml`

* New top-level `sessionRecording` section (the recommended way):
  * `enabled` — turns on worker capture (`rec_*.mkv` on the shared volume)
    and dispatcher upload together.
  * `quality` — `144p|240p|360p|480p|720p|1080p` (drives ffmpeg CRF /
    framerate on the worker).
  * `upload.*` — S3 bucket / prefix / region / creds / endpoint /
    compress / SSE / `uploaderStatePath`.
* `dispatcher.config.recording` and `webWorker.config.recording` shrunk
  from full default blocks to `{}` with override examples in the
  comments. **This is the bugfix** — when those blocks shipped non-empty
  by default, the dispatcher template took them as a service override and
  silently overrode `sessionRecording.upload.*` whenever an operator
  enabled recording via the unified section.
* `sessionRecording` docstrings describe the new on-disk contract: worker
  writes `rec_*.mkv`, dispatcher remuxes to `rec_*.mp4` with `ffmpeg`
  before S3 upload — the customer-facing artifact in S3 stays `.mp4`.

### `templates/dispatcher-deployment.yaml`

Rewrites the recording-env precedence to **per-field**:

1. Read unified `sessionRecording.upload.<field>` (if set).
2. Overlay `dispatcher.config.recording.<field>` (if set, per field —
    via `hasKey` for booleans / `if $r.<field>` for strings).
3. Built-in defaults otherwise.

This mirrors the pattern already used by `web-worker-deployment.yaml` and
removes the trap where an empty/legacy override block always won. Boolean
env vars are rendered with `ternary | quote` to avoid the
`json: cannot unmarshal bool into …EnvVar.value` Helm error we hit on
older clusters.

### `templates/web-worker-deployment.yaml`

* Comment clarifies the per-field precedence.
* Splits the old `if/else` into two passes (read unified, then overlay
  per-field overrides) so the worker template is structurally identical
  to the dispatcher template.
* Same boolean-quoting fix as the dispatcher template.

### `README.md`

* Replaces the inaccurate "Partial Recordings" guidance (which used to
  say "increase staleGraceSeconds or accept truncated recordings") with
  the new MKV-on-disk + SIGTERM-flush + dispatcher-remux story.
* Drops the dangling reference to the (relocated) `BACKWARD_COMPAT_TEST.md`.
* Documents the `session_recorder` preflight (`/etc/shared` writable,
  `/bin/ffmpeg` executable) so operators recognise the fail-fast log line.

### `Chart.yaml`

* `version: 3.1.0 → 3.2.0` (new top-level value section + bugfix).

## Verification — `helm template` matrix

| Scenario | `ENABLE_RECORDING` (worker) | `ENABLE_RECORDING_UPLOAD` (disp.) | `RECORDING_S3_BUCKET` |
|---|---|---|---|
| **defaults** | `"false"` | `"false"` | `""` |
| **unified only** (`sessionRecording.enabled=true`, `sessionRecording.upload.s3Bucket=unified-bucket`) | `"true"` | `"true"` | `"unified-bucket"` ← was `""` before fix |
| **unified + per-field override** (`dispatcher.config.recording.s3Bucket=override-bucket`, `webWorker.config.recording.quality=720p`) | `"true"`, quality `"720p"` | `"true"` | `"override-bucket"` |

`helm lint` clean.

## Backward compatibility

Existing deployments using `dispatcher.config.recording.*` and
`webWorker.config.recording.*` continue to work — those fields still
override the unified section per-field. The only behavioural change is
that an **untouched** values.yaml no longer has a non-empty
`dispatcher.config.recording` block silently overriding
`sessionRecording.upload.*`. Operators who want the old behaviour can
re-add the explicit fields under `dispatcher.config.recording.*`.

## Test plan

* [ ] `helm lint charts/akeyless-zero-trust-web-access` (✅ done)
* [ ] `helm template` against the three scenarios above (✅ done — matrix
      in this PR)
* [ ] Install `3.2.0` on the k3s sandbox with new images from the
      linked `akeyless-miscellaneous` PR; trigger a session; verify
      playable `.mp4` lands in S3.
* [ ] Verify upgrade from `3.1.0` (where applicable) does not break
      existing deployments using legacy per-service config.

## Linked PR

* `akeyless-miscellaneous`:
  [ASM-17214-session-recording-support-for-s3upload-quality-compression](https://github.com/akeylesslabs/akeyless-miscellaneous/pull/115)
  (new `session_recorder` Go service, MKV-on-disk pivot, dispatcher remux,
  uploader hardening).

## Refs

* Jira: ASM-17214

## Summary by CodeRabbit

* **New Features**
  * Unified optional session recording with cloud upload, configurable quality, compression, SSE/KMS, uploader state path, and per-service override precedence
  * Worker-side recording toggles and dispatcher-controlled upload settings; conditional credential handling

* **Documentation**
  * Expanded session recording guide: setup, single vs multi-node warning, precedence rules, upload behavior, watchdogs, troubleshooting, and examples

* **Chores**
  * Chart version bumped; validator resources made configurable; worker liveness probe updated to exec script

### DIFF
--- a/charts/akeyless-zero-trust-web-access/values.yaml
+++ b/charts/akeyless-zero-trust-web-access/values.yaml
@@ -351,9 +351,28 @@ webWorker:
   #   limits:
   #     memory: 4Gi
 
+  # Two layers of session-cleanup defence on the worker:
+  #
+  # 1. The web_worker_agent itself removes /tmp/occupied as soon as `/healthy`
+  #    detects that the VNC client closed the connection (clearSessionMarkers()
+  #    in handlers/register_handlers/server.go). This is the fast path — no
+  #    pod recycle is needed and the worker re-enters the dispatcher pool on
+  #    the next /healthy poll (~30s after the user closes the tab).
+  # 2. The liveness probe below is the *safety net* for the case where the
+  #    agent process itself dies. Once `/tmp/occupied` is older than
+  #    sessionCleanup.staleGraceSeconds AND no client is on :5800, the probe
+  #    wipes the markers and signals kubelet to recycle the pod.
   sessionCleanup:
+    # How long to wait after `/tmp/occupied` is created before the probe is
+    # allowed to declare the worker stranded. Must cover the worst-case
+    # browser-mount latency (token exchange + Firefox render). 30s is enough
+    # for a healthy network; raise if your sessions need >30s warm-up.
     staleGraceSeconds: 30
 
+  # Tighter defaults than the chart used to ship with: failureThreshold * periodSeconds
+  # is the upper bound on how long a dead-agent pod stays stranded. With these values,
+  # recycle happens within ~30 seconds of the third consecutive failure, vs. the older
+  # 60 * 30s = 30-minute window that hid the ASM-17214 lifecycle regression.
   livenessProbe:
     initialDelaySeconds: 60
     periodSeconds: 10
--- a/charts/akeyless-zero-trust-web-access/values.yaml
+++ b/charts/akeyless-zero-trust-web-access/values.yaml
@@ -77,6 +77,30 @@ sessionRecording:
     # Default: upload/recording_uploader_state.json (under /etc/shared)
     uploaderStatePath: ""
 
+  # Worker-side lifecycle watchdog configuration
+  # These knobs cap a single recording's wall-clock duration and detect
+  # "browser tab opened but client never connected" abort scenarios.
+  # The defaults are safe for typical interactive sessions; extend
+  # maxDurationSeconds for environments with very long-running screens
+  # (e.g. 24/7 dashboards). Empty string keeps the in-image defaults.
+  watchdog:
+    # WORKER_CLIENT_CONNECT_TIMEOUT_SECONDS — how long the worker waits for
+    # the browser to actually establish a websocket on :5800 after the
+    # session is assigned. If the client never appears, the worker tears
+    # down the session so the liveness probe can recycle the pod.
+    # In-image default: 90 seconds.
+    clientConnectTimeoutSeconds: ""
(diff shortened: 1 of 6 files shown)


lidorb-akeyless (approved):
nice, see if you can add resource limits to the Validator like Cisco asked with this release?
It will save us releasing a dedicated version later

lidorb-akeyless (approved):

charts/akeyless-zero-trust-web-access/values.yaml
```diff
@@ -29,6 +29,78 @@ httpProxySettings:
...
+    s3Prefix: ""
+    
+    # Optional custom endpoint for S3-compatible services (MinIO, Wasabi, NetApp StorageGRID, etc.)
+    # Leave empty for AWS S3
+    s3Endpoint: ""
+    
+    # S3 credentials (NOT RECOMMENDED: use secretKeyRef in dispatcher.env or IAM/IRSA instead)
+    # Leave empty to use AWS default credential chain (instance profile, environment, etc.)
+    s3AccessKeyId: ""
+    s3AccessKeySecret: ""
```
omriezra: don't let passing plain secret, optional with secret name and relevant key refs

Hanan-Akeyless: done

omriezra (approved):

PR #391, opened 2026-06-26
ASM-18712 | fix hardcoded gw tls cert in sra deployments

kgal-akl:
## Summary by CodeRabbit

* **New Features**
  * Made the TLS certificate secret key name configurable for both secure remote access and SSH deployments, with a default of `tlsCertificate`.
  * This improves compatibility with TLS secrets that use custom key structures.
* **Chores**
  * Bumped the Helm chart version for the gateway from `3.1.4` to `3.1.5`.

### DIFF
--- a/charts/akeyless-gateway/templates/akeyless-secure-remote-access/deployment.yaml
+++ b/charts/akeyless-gateway/templates/akeyless-secure-remote-access/deployment.yaml
@@ -98,13 +98,6 @@ spec:
             - name: GATEWAY_AUTHORIZED_ACCESS_ID
               value: {{ .Values.globalConfig.authorizedAccessIDs | quote }}
             {{- end }}
-            {{- if .Values.globalConfig.TLSConf.tlsExistingSecret }}
-            - name: AKEYLESS_GW_CERTIFICATE
-              valueFrom:
-                secretKeyRef:
-                  name: {{.Values.globalConfig.TLSConf.tlsExistingSecret | quote  }}
-                  key: {{ (.Values.globalConfig.TLSConf).tlsCertificateSecretKeyName | default "tlsCertificate" }}
-            {{- end }}
             - name: REMOTE_ACCESS_TYPE
               value: "web"
             {{- if .Values.globalConfig.gatewayAuth.azureObjectID }}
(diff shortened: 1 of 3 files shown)


charts/akeyless-gateway/templates/akeyless-secure-remote-access/deployment.yaml
```diff
@@ -98,13 +98,6 @@ spec:
             - name: GATEWAY_AUTHORIZED_ACCESS_ID
               value: {{ .Values.globalConfig.authorizedAccessIDs | quote }}
             {{- end }}
-            {{- if .Values.globalConfig.TLSConf.tlsExistingSecret }}
-            - name: AKEYLESS_GW_CERTIFICATE
```
kgal-akl, line 102: this env var is some leftover/dead code as it doesn't show up in ztb code:

```bash
grep "AKEYLESS_GW_CERTIFICATE" -rnw ~akl_ztb | wc -l
       0
```

Hanan-Akeyless: not leftover, it's used in the bastions 
Image

kgal-akl: @Hanan-Akeyless

**the env var `AKEYLESS_GW_CERTIFICATE` is rendered in the helm chart. however, if you review the ztb code, you will see that this env var is not actually read anywhere.** 

i did some more research and found that the ztb repo has a remote branch named `AKEYLESS_GW_CERTIFICATE` that [includes changes](https://github.com/akeylesslabs/zero-trust-bastion/compare/master...AKEYLESS_GW_CERTIFICATE) to read that env var and use it as a root CA for sdk's http client. however, this branch has not actually been merged so the env var is not actually read in neither bastion. 
the way the CAs are feteched today is using `pkg/certstore`. both web and ssh bastions call `certstore.NewManager(os.TempDir())` followed by `BootstrapFromGateway` ([source](https://github.com/akeylesslabs/zero-trust-bastion/blob/master/bastions/web/main.go#L220-L221))

PR #396, opened 2026-06-26
ReferenceGrant for cross-namespace routing [5/6]

drzln:
Part **[5/6]** of the ** — Kubernetes Gateway API support** stack. Stacked on #395.

Gateway API denies cross-namespace `route ↔ Gateway` and `route ↔ backend` references by default; a **ReferenceGrant** in the target namespace is the explicit, auditable grant. This is exactly the shape from the ticket's reference example (Vermeer / shared-Gateway scenarios).

```yaml
gatewayAPI:
  referenceGrants:
    - name: allow-shared-gateway-to-akeyless
      # namespace: defaults to the release namespace
      from:
        - { group: gateway.networking.k8s.io, kind: HTTPRoute, namespace: gateway }
      to:
        - { group: "", kind: Service }
```

`templates/referencegrant.yaml` → one `ReferenceGrant` (`gateway.networking.k8s.io/v1beta1`) per entry; `from`/`to` passed through verbatim; created in the release namespace unless `namespace` is set. Schema enforces `required: [name, from, to]`.

## Render tests
`tests/gateway_api_referencegrant_test.yaml` — 3 cases (disabled-by-default, full v1beta1 render with verbatim from/to, explicit target namespace).

Verified locally: `helm unittest` **30/30** across all 5 suites ✅, `helm lint` ✅. Chart `1.65.0 → 1.66.0`.

### DIFF
--- a/charts/akeyless-api-gateway/Chart.yaml
+++ b/charts/akeyless-api-gateway/Chart.yaml
@@ -16,7 +16,7 @@ maintainers:
 type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 1.65.0
+version: 1.66.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: 4.53.0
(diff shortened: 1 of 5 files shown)


drzln:
Closing — moving this work to the private akeyless-environments repo. The Gateway API surface for the akeyless gateway belongs in the private helm, not the public OSS chart. Tracked internally.

PR #398, opened 2026-06-27
Gateway API support on the akeyless-gateway chart (alongside Ingress)

drzln:
Adds **Kubernetes Gateway API** support to the customer-facing **`akeyless-gateway`** chart — the chart external customers use to deploy the gateway — **alongside** its existing Ingress, so customers can migrate ahead of the [NGINX Ingress retirement (Mar 2026)](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/).

> This is the correct home per @kobbi: the **public `akeylesslabs/helm-charts`** chart **`akeyless-gateway`** (not the legacy `akeyless-api-gateway`, not a private wrapper). It supersedes the earlier private `akeyless-environments` PRs and the old public draft stack.

## What it adds (opt-in, default OFF)
A top-level `gatewayAPI:` block + `templates/gateway-api/*`:
- `Gateway` + `HTTPRoute` (v1) · `TLSRoute` + `TCPRoute` (v1alpha2) · `ReferenceGrant` (v1beta1) · HTTP→HTTPS redirect · `extraObjects`
- Controller-agnostic (`gateway.gatewayClassName`: cilium/nginx/istio/aws-alb/gke)
- AWS ALB: `gateway.infrastructure.parametersRef` (`LoadBalancerConfiguration`) + ACM hostname auto-discovery (omit `certificateRefs`)
- `gateway.addresses` (GKE/AWS static IPs); routes target the chart's own Service on named ports (`gateway`/`kmip`/`hvp`), with a per-route `serviceName` override for the separate SRA SSH service
- Fail-loud guards **G1–G5** (G1 = `gatewayAPI` ⊕ the chart's `gateway.ingress`)

**Defaults OFF — existing installs are unaffected** (no gateway-api objects render unless `gatewayAPI.enabled=true`).

## Tested
- **26 helm-unittest** cases (`tests/gateway-api-*`) — Gateway/listeners, route port-resolution incl. SSH `serviceName` override, TLS terminate + passthrough, redirect (301), ReferenceGrant, extraObjects, composition matrix, and a `failedTemplate` per guard.
- `helm lint` clean; render verified (backendRefs resolve to the chart fullname on port 8000).
- The identical Gateway API surface was proven **live end-to-end** during development (kind + Envoy Gateway + MetalLB, real L7/L4 traffic) — this PR is that surface ported into the source chart.
- The repo's `ct` workflow install-tests the chart on every change.

[...]

PR #399, opened 2026-06-27
[DEFERRED] Gateway API Layer-2 — optional HTTPRoute enhancements (weighted backends, RequestMirror, rich matches, cross-ns backendRef)

drzln:
**Deferred, not blocked.** #420 ships the Tier 1 delivery (Gateway + HTTPRoute + redirect, strict `ingress.yaml` parity) as the thing that actually closes. This PR holds the fuller HTTPRoute surface -- weighted/canary backends, RequestMirror, rich matches (path+header+method+query), cross-namespace backendRef, and the ReferenceGrant it depends on -- as an intentionally separate, optional follow-up.

Checked against the actual `ingress.yaml` template: none of this has parity there (single host/path/backend only, and core Ingress can't do cross-namespace backends at all), so none of it is required to close the ticket. All of it is additive/opt-in/default-off already -- Layer-1 (Tier 1) render is byte-identical when these fields are unset -- so nothing here is blocking anyone; it's real, tested capability sitting ready for whenever there's a concrete need to enable it.

Kept as its own draft PR (not merged into #420, not deleted) so it can be delivered iteratively -- reviewed and picked up on its own timeline, layered on top of #420 once that lands.

Original description below, unchanged.

---

**Optional completeness layer** stacked on #398. Extends the `akeyless-gateway` gatewayAPI surface to the **full Gateway API HTTPRoute** capability set — **additive, opt-in, default OFF** (Layer-1 render is byte-identical when the new per-route fields are unset).

## Adds (per route)
- **Rich matches** — path **+ header + method + query**
- **Filters** — URLRewrite · request/response header modifiers · request mirror
- **Weighted backends** — canary split, with named-port resolution
- **Cross-namespace backendRef** — pairs with the ReferenceGrant
- **Per-route timeouts**

`G5` servicePort guard is now `backendRefs`-aware.

## Tested
- **34 helm-unittest** cases (+8 Layer-2), comment-free.
- The identical Layer-2 surface was proven **live end-to-end** during development (kind + Envoy Gateway): URLRewrite path-rewrite, header-match (hit + 404 miss), request/response header modifiers, weighted split, and cross-namespace data-plane.

Chart `3.2.0 → 3.3.0`. Merge after #398.


drzln:
## Validated live on Cilium — Vermeer's Gateway API controller

All earlier integration runs used **Envoy Gateway**. Since Vermeer runs **Cilium** as its Gateway API controller, the chart's Gateway API surface is now also validated against Cilium in an automated kind harness (`charttest integration`, the keyway tool).

**Automated scenario `gateway-api-cilium` — 8/8 green** (fresh provision → render → apply → assert → teardown), rendering **this chart**:

```
PASS gateway-api-cilium  cluster             kind cluster ready (CNI-less)
PASS gateway-api-cilium  gateway-api-crds    Gateway API CRDs installed (experimental v1.1.0)
PASS gateway-api-cilium  controller          daemonset/cilium ready; GatewayClass cilium
PASS gateway-api-cilium  loadbalancer        LB pool from kind subnet (Cilium L2 IPAM)
PASS gateway-api-cilium  vermeer-cilium-l7/stub-backend       1 backend service ready
PASS gateway-api-cilium  vermeer-cilium-l7/render-apply       2 Gateway API objects applied
PASS gateway-api-cilium  vermeer-cilium-l7/gateway-programmed Gateway Programmed=True
PASS gateway-api-cilium  vermeer-cilium-l7/route              HTTPRoute Accepted+ResolvedRefs
```

**Recipe:** CNI-less kind (`disableDefaultCNI`, `kubeProxyMode: none`) + Cilium 1.16.5 (`kubeProxyReplacement`, `gatewayAPI.enabled`, `l2announcements`) + Cilium's own `CiliumLoadBalancerIPPool` for the Gateway address. The chart renders a `gatewayClassName: cilium` Gateway + HTTPRoute, Cilium **programs the Gateway** (`Programmed=True`) and **resolves the route** (`Accepted+ResolvedRefs`).

**No chart change was needed for Cilium** — `gatewayClassName` is already a first-class value (the helm-unittest suite renders against `cilium` by default), so switching controllers is pure configuration. Live data-plane traffic through Cilium (Gateway LB IP → backend `/hostname`; wrong-host → 404) was additionally confirmed manually; the automated scenario asserts at the condition level because Cilium's per-Gateway proxy Service exposes no pod endpoints for `kubectl port-forward`.

drzln:
Test evidence as of `411172c`: 41/41 helm-unittest passing (base gateway suite + guards + Layer-2 HTTPRoute surface + matrix + routes). Merged in #398's latest (the 4 override-gap tests) cleanly, no conflicts. CI `chart-test` green on this commit.

stanislav-akeyless (approved):

drzln:
Retargeted base from the closed #398's branch to `main` — this PR now carries the full combined delivery: Tier 1 (Gateway + HTTPRoute + redirect + ReferenceGrant, ex-#398/#417) merged with Layer-2 (rich HTTPRoute matching — path/header/method, filters, weighted/cross-namespace backends, per-route timeouts). Rationale: #417 (a fresh duplicate of Tier 1 alone, opened because #398's branch got a stuck GitHub Actions integration) hit a separate, still-unresolved platform issue — brand-new PR opens aren't triggering check-suites in this repo right now, while pushes to already-open PRs (like this one) do. Rather than wait on an unpredictable platform recovery, landing everything here gets the whole thing green today: 32/32 helm-unittest, `helm lint` clean, `chart-test` CI green (https://github.com/akeylesslabs/helm-charts/actions/runs/29505454334). Re-requested review from @EldadH89 and @stanislav-akeyless since the diff scope materially changed from what was previously approved. Closing #417 as superseded — its content is a strict subset of what's here.

Alex-Souslik (commented):
Thanks for the thorough work here — the guard rails (G1–G5), the `Layer-1 unchanged` regression test, and the `ct list-changed` fix (the `>>> ` debug-trace leak into `$CHANGED_CHARTS`) are all solid.

On the Layer-2 surface, I'd like to talk through scope before this merges, mainly because it's public/customer-facing and I want us to be deliberate about what we're committing to support long-term.

The motivating driver here (per #398) is giving nginx-ingress customers a migration path ahead of the retirement. Given that, I went back to `ingress.yaml` to see what parity actually requires: it renders host + path + one backend on a named port, plus TLS — nothing else is templated (annotations are a raw passthrough, so anything controller-specific has always been the customer's own business, not something this chart parameterizes).

A few of the Layer-2 additions go further than that baseline, and I couldn't find a documented case for them:
- **`ReferenceGrant` + cross-namespace `backendRef`s** — don't see this pattern used elsewhere in the chart or docs.
- **Weighted/multi `backendRefs` (canary/blue-green)** — the chart's only supported rollout mechanism today is `Deployment.strategy` (rolling update); no existing canary/blue-green story to extend.
- **`RequestMirror` filter** — same shape of concern, no tied use case.
- **Rich `matches` (header/method/query) + `URLRewrite`/header-modifier filters** — I initially thought these might matter for nginx-annotation parity (rewrite-target, canary-by-header), but our [SRA sticky-sessions doc](https://docs.akeyless.io/docs/sra-sticky-sessions-and-ingress-patterns) — the one place we do publish ingress-layer guidance — points customers at controller-specific annotations/policies for that, not at anything a chart abstraction would replace. Our established pattern for controller-specific needs seems to be "customer self-serves it," not "the chart parameterizes it."

Would you be open to scoping the mergeable version down to: optional `Gateway` (so a customer gets a working instance without a pre-existing shared Gateway) + `HTTPRoute` (hostname, path match, one backend on the chart's existing named ports, TLS + redirect) — a fairly literal mirror of `ingress.yaml`? That solves the nginx-retirement migration need without expanding what this chart's job is.

Genuinely open to being wrong if there's a concrete customer ask behind any of the above — just want to make sure we're building it because someone needs it, not because Gateway API makes it possible. The rest of the engineering here is solid.

Two small unrelated nits:
- `values.yaml` has two stray whitespace-only changes (`{{ .Release.Name }}` → `{{.Release.Name }}`) in `cacheHA` that look unintentional.
- Chart version jumps `3.1.9 → 3.3.0` (skipping 3.2.0) — flagging in case that wasn't intentional.

PR #409, opened 2026-07-07
tests for the HTTPRoute-only Gateway API slice

drzln:
**Validates #408 (HTTPRoute-only Gateway API slice).** Stacked on top of it — merge #408 first, this rebases onto `main` once it does.

Coverage (helm-unittest, 20/20 green):
- **Guards** — G1 (gatewayAPI vs ingress mutual exclusion), G2 (parentRefs required, httproute.yaml path), G3 (parentRefs required, gatewayApi.validate path), G4 (unknown servicePort fails loud with valid names listed), G5 (quoted numeric servicePort resolves)
- **Routes** — HTTPRoute → gateway Service:8000; two routes get correctly-indexed, distinct names *and* distinct hostnames (`documentSelector`-based, one assertion per route — not just a document count); httproute-redirect.yaml (disabled by default, inherits the top-level parentRefs + default sectionName http, explicit parentRefs override)
- **Advanced route matching** — rich matches (path/header/method), URLRewrite, request/response header modifiers, RequestMirror, weighted/canary backends, cross-namespace backendRef, per-route timeouts, confirms the minimal (single path/single backend) shape is unaffected

No Gateway-resource test coverage — Gateway creation isn't part of this slice; every route attaches to an externally supplied `gatewayAPI.parentRefs`.

No TLSRoute/TCPRoute/ReferenceGrant test cases — that surface isn't part of the #408 slice.

**Note on `documentIndex` vs `documentSelector`:** the installed helm-unittest (v1.0.3) does not correctly scope an `equal` assertion by `documentIndex` against this chart's `range`-generated multi-document `httproute.yaml` template — verified with an isolated single-assertion repro (a lone `documentIndex: 1` assertion was evaluated against document 0's content). `documentSelector` (content-based targeting) works correctly and is used instead for the multi-route case.

Ref: (https://akeyless.atlassian.net/browse/

 : https://akeyless.atlassian.net/browse/ ?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Gateway API routes now support advanced matching options, including URL rewrites, header modifications, request mirroring, weighted backends, cross-namespace targets, and per-route timeouts.
  - HTTPS redirect routes support inherited or custom parent references and 301 redirects.
  - Multiple routes receive deterministic, unique names and hostname suffixes.

[...]

PR #408, opened 2026-07-07
Kubernetes Gateway API support (HTTPRoute only) on akeyless-gateway

drzln:
**Scoped deliverable: HTTPRoute only.**

Adds `gatewayAPI.*` to the `akeyless-gateway` chart:
- `HTTPRoute` — path/header/method matches, filters (URLRewrite, header modifiers, RequestMirror), weighted/cross-namespace `backendRefs`, per-route timeouts, HTTP→HTTPS redirect
- Attaches to an **existing** Gateway via required `gatewayAPI.parentRefs` — this chart does not create/own a Gateway resource
- Fail-loud guard: `gatewayAPI.enabled` and `gateway.ingress.enabled` are mutually exclusive
- Default `gatewayAPI.enabled: false` — byte-identical output when disabled

**Intentionally out of scope for this slice:** Gateway creation/ownership, TLSRoute, TCPRoute (KMIP/HVP/SSH), ReferenceGrant, `extraObjects`. See "Follow-up / optional layers" below for where each lives.

**No tests in this PR by design.** Test coverage (helm-unittest suites) lands in a follow-up branch/PR stacked on this one, so this change can be validated independently before test authoring is reviewed.

Ref: (https://akeyless.atlassian.net/browse/

## Follow-up / optional layers

This PR is the current deliverable per (https://akeyless.atlassian.net/browse/ (the "attach HTTPRoutes to a Gateway the customer already runs" option). Everything below layers optional complexity on top — none of it blocks merging this PR, and which of these ship is a customer-facing product decision, not a technical prerequisite.

**Stacked directly on this PR:**
- #409 — helm-unittest coverage for this HTTPRoute-only slice
- #410 — CI Cilium conformance (kind + Cilium Gateway API controller) for this slice

**Alternative/richer design — the chart creates and owns the Gateway itself** (rather than attaching to one the customer already runs): #398 + #399, with their own conformance/CI stack (#401, #402, #403). This track additionally covers TLSRoute (SNI passthrough), TCPRoute (KMIP/SSH raw-L4), ReferenceGrant (cross-namespace), `extraObjects` (per-cloud policy CRs — AWS ALB `LoadBalancerConfiguration`/`TargetGroupConfiguration`, GKE `GCPGatewayPolicy`), and a richer HTTPRoute match/filter surface (Layer-2 in #399: URLRewrite, header injection, canary weighting, cross-namespace backends, timeouts).

[...]

PR #410, opened 2026-07-07
CI Cilium conformance for the HTTPRoute-only slice

drzln:
**Real Cilium proof for #408, kept out of it.** Stacked on #409 (helm-unittest), which is stacked on #408 (the HTTPRoute-only implementation).

Adds a GitHub Actions job that:
1. Spins up a kind cluster (kube-proxy replacement, no default CNI)
2. Installs **Cilium 1.19.5** as the Gateway API controller (`gatewayAPI.enabled=true`), waits for `GatewayClass cilium` to reach `Accepted`
3. Applies a fixture `Gateway` (`shared-gw`, `chart-testing`) standing in for a platform team's pre-existing Gateway, waits for it to reach `Programmed=True`, then installs the `akeyless-gateway` chart pointed at it via `gatewayAPI.parentRefs` (this slice attaches HTTPRoutes to an existing Gateway, it does not create one)
4. Runs a `helm test` conformance pod that waits for the fixture `Gateway` to reach `Programmed=True` and the chart's `HTTPRoute` to reach `Accepted` + `ResolvedRefs=True`

This is the same mechanism previously proven 8/8 on the broader `gwapi-cilium-dataplane` branch, re-scoped here to exactly what #408 ships (no chart-owned Gateway, `parentRefs` always required). **No data-plane echo scenario** in this PR: that scenario depends on `extraObjects`, which isn't part of the HTTPRoute-only slice (would be scope creep back into #408). If we want live-traffic-through-Cilium proof too, that's an easy follow-up once `extraObjects` is (if ever) added to this chart's scope.

**Fixed a real cross-job regression during review:** the fixture `Gateway` object initially lived in `.github/chart-fixtures/akeyless-gateway/`, which the pre-existing, **required** `chart-test` job also applies wholesale on a plain kind cluster with no Gateway API CRDs installed — that broke `chart-test` with `no matches for kind "Gateway"` (confirmed by comparing that job's status on the commit before/after the fixture was added: green → red, same job). Moved the fixture to `.github/gateway-api/fixtures/`, consumed only by this workflow (which installs the Gateway API CRDs itself).

Merge order: #408 → #409 → this.

Ref: (https://akeyless.atlassian.net/browse/

 : https://akeyless.atlassian.net/browse/ ?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Added a reusable GitHub composite action to set up Gateway API on a kind cluster with Cilium, including configurable Gateway API/Cilium versions and optional dataplane/L2 behaviors.
  - Added Gateway API test configuration for Cilium, including shared Gateway fixtures and a tailored kind cluster configuration.
  - Added a Helm-rendered Gateway API conformance test validating programmed gateways and HTTPRoute acceptance.

- **Tests**
  - Introduced a “Gateway API Test” workflow that provisions a kind-based environment and collects diagnostics on failure.
  - Extended chart CI to run `helm-unittest` for changed charts.


drzln:
Closing as superseded — #417 (gatewayAPI.gateway.create toggle) is a strict superset of #408's HTTPRoute-only/attach-to-existing-Gateway scope, so this Cilium-conformance coverage for #408 is redundant. If #417 needs the same live-Cilium conformance proof, it'll be added there directly rather than kept as a separate stacked PR.

PR #412, opened 2026-07-08
[ASM-18793] ZTWA v2: split proxy

yogevk-akeyless:
split proxy into its own Deployment/Service + compose service

Add webProxy Deployment/Service (LoadBalancer :19414) served by the new zero-trust-web-proxy image; remove :19414 from the dispatcher Deployment/Service; add webProxy values + schema; update compose; major chart bump (4.0.0) + V2 migration note (customers repoint their proxy endpoint).

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated web proxy Deployment and Service (with health checks), plus optional Gateway API TCPRoute support.
  * Updated Docker Compose to run proxy traffic via a separate proxy service on port `19414`.
* **Breaking Changes**
  * Dispatcher no longer exposes proxy traffic; proxy configuration and exposure moved to the dedicated `-proxy` endpoint.
  * Dispatcher ports were updated (proxy port removed) and its Service now exposes only the HTTP port.
* **Documentation**
  * Expanded chart docs and upgrade notes to reflect the moved proxy settings.
  * Bumped the Helm chart version to `4.0.0`.


charts/akeyless-zero-trust-web-access/templates/NOTES.txt
```diff
@@ -30,3 +30,26 @@ This release supports IPv4, dual-stack, and IPv6-only clusters (e.g. EKS in IPv6
 - Rollback warning: on IPv6-only / IPv6-primary clusters, do NOT roll back to an
   older ZTWA version — earlier images bind IPv4-only and fail readiness on the IPv6
   pod IP with no workaround.
+
+Akeyless Zero Trust Web Access — V2 migration (BREAKING)
```
yogevk-akeyless: its became a long scroll. we need to see what w do with this content.
@lidorb-akeyless @Hanan-Akeyless

lidorb-akeyless: we don't need to expend this for every version or braking change.
only when manual action is required

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
```diff
@@ -0,0 +1,136 @@
...
+        app: web-proxy
+    spec:
+      securityContext:
+        # Match the runtime user (nobody). The dispatcher's fsGroup 10000 is its
+        # shared-upload PVC group; the proxy mounts no such volume.
+        fsGroup: 65534
+      volumes:
+        # Shared read-write config dir. The proxy needs zero privileged creds, so the
+        # dispatcher's privileged-access secret and -cm-akeyless configmap are NOT mounted.
+        - name: ng-config
```
lidorb-akeyless, line 35: why do you need this mounted?
who consumes this?

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
```diff
@@ -0,0 +1,136 @@
...
+            allowPrivilegeEscalation: false
+          ports:
+            - containerPort: 19414
+              name: proxy
+          env:
+{{- if .Values.httpProxySettings.http_proxy }}
+            - name: HTTP_PROXY
+              value: {{ .Values.httpProxySettings.http_proxy }}
+            - name: http_proxy
+              value: {{ .Values.httpProxySettings.http_proxy }}
```
lidorb-akeyless, line 101: why the duplication of env? (both inside the if, and you have the same if twice in a row)

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
```diff
@@ -0,0 +1,136 @@
...
+{{- end }}
+{{- if .Values.httpProxySettings.https_proxy }}
+            - name: HTTPS_PROXY
+              value: {{ .Values.httpProxySettings.https_proxy }}
+            - name: https_proxy
+              value: {{ .Values.httpProxySettings.https_proxy }}
+{{- end }}
+{{- if .Values.httpProxySettings.no_proxy }}
```
lidorb-akeyless, line 109: new mode?

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
yogevk-akeyless: ng-config is just the volume's name (an emptyDir, named after the dispatcher's equivalent). The proxy reads its config from `/etc/akeyless` — the init container writes 2 files there: the gateway URL and the allowed access-IDs. `web_proxy.sh` reads both at startup, and squid's auth script checks every login against the access-ID list. Same pattern as the dispatcher.

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
yogevk-akeyless: Copied from the dispatcher — it was already like that. The two spellings are needed: curl reads only `http_proxy`, the Go CLI reads `HTTP_PROXY`.

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
yogevk-akeyless: No, this is the existing `httpProxySettings` (corporate egress proxy), same as on the dispatcher and worker. Not related to proxy mode.

ref: main's dispatcher-deployment.yaml:171-188

p.s - 
naming is confusing is here **httpProxySettings** is not about our web proxy, it's the customer's corporate egress proxy, which this pod's own outbound calls (CLI auth, gateway /status) must go through in locked-down networks. Same as dispatcher/worker.

charts/akeyless-zero-trust-web-access/Chart.yaml
```diff
@@ -16,7 +16,9 @@ maintainers:
 type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 3.4.1
+version: 4.0.0
+# appVersion is set to the released ZTWA image tag at release time (must match the
+# zero-trust-web-{dispatcher,worker,proxy} images)
 appVersion: 2.0.1
```
Hanan-Akeyless, line 22: `2.0.1` is already released
we need to coordinate this change

1. merging [Asm 18793 separate proxy backend](https://github.com/akeylesslabs/akeyless-miscellaneous/pull/147) --> no helm release!
2. decide the new image tag (major/minor/patch) --> release ZTWA with tag `T`
3. put tag `T` here as well before merging the chart
4. merge the chart --> will auto release it

charts/akeyless-zero-trust-web-access/templates/proxy-deployment.yaml
```diff
@@ -0,0 +1,136 @@
+{{- $replicas := .Values.webProxy.replicaCount }}
+{{- if kindIs "invalid" $replicas }}
+{{- $replicas = .Values.dispatcher.replicaCount }}
+{{- end }}
```
Hanan-Akeyless, line 7: When `HPA.enabled=true`, the dispatcher Deployment omits replicas entirely (HPA owns it) and `dispatcher.replicaCount` becomes inert for the dispatcher - but it still statically sizes the proxy via the inheritance fallback. The proxy has no HPA of its own. 

**Why it matters:** An operator scaling via HPA reasonably assumes `dispatcher.replicaCount` is dead config, it silently pins the proxy fleet size instead. Surprising coupling, hard to debug from symptoms ("why do I have 3 proxies?"). 

**Suggested fix:** Cheapest: extend the values.yaml comment - "when HPA is enabled, set webProxy.replicaCount explicitly, the dispatcher.replicaCount fallback does not track HPA." 

**Better:** under HPA.enabled, require/default an explicit webProxy.replicaCount.

yogevk-akeyless: done

charts/akeyless-zero-trust-web-access/Chart.yaml
yogevk-akeyless: I will do that on the main branch + add the comment there just in case

PR #417, opened 2026-07-16
Gateway API support on the akeyless-gateway chart — Tier 1 (Gateway + HTTPRoute + redirect + ReferenceGrant)

drzln:
Replaces #398 — identical content (`1205136`), pushed under a fresh branch because GitHub Actions got stuck and never created a check-suite for #398's last 3 pushes (confirmed via 6 separate remedies: 3 fresh pushes, close/reopen, an empty retrigger commit, and a workflow disable/enable cycle — none produced a `github-actions` check-suite on that branch).

## Scope (Tier 1, per the narrowed definition)
Gateway (create or attach-to-existing via `gatewayAPI.gateway.create`) + HTTPRoute + HTTP→HTTPS redirect + ReferenceGrant (kept, and here's why: cross-namespace `backendRef`s are silently ignored by every Gateway API implementation without a matching ReferenceGrant — dropping it would silently break any HTTPRoute that targets a Service in another namespace).

TLSRoute / TCPRoute (KMIP/HVP/SSH) / `extraObjects` were removed from this PR's scope — Tier 1.5/2, tracked separately, not deleted from the org's plans, just not in this slice.

## Relationship to #408/#409/#410
#408 ("HTTPRoute-only, attach to existing Gateway") is a **subset** of this PR's scope now that `gatewayAPI.gateway.create` unifies both the chart-owns-the-Gateway and attach-to-existing paths in one toggle (see #398's own history — the toggle was added and fully tested, 33/33, before this Tier-1 narrowing). Once this PR is green and merged, #408/#409/#410 are redundant and should be closed as superseded — flagging for confirmation before doing that since #408 already carries a human APPROVED review.

## Tested
24/24 helm-unittest (`gateway-api-{gateway,guards,matrix,routes}_test.yaml`), `helm lint` clean. Verified via `helm template` that `tls.mode=Passthrough` fails loudly (Tier 1.5 guard) and that a minimal render has zero dangling TCP/TLS content.

## Summary by CodeRabbit

- **New Features**
  - Added Gateway API support for the Akeyless Gateway Helm chart.
  - Added configurable Gateways, listeners, TLS termination, HTTPS redirects, HTTP routes, backend services, and cross-namespace reference grants.
  - Added attach mode for connecting routes to an existing Gateway.
  - Added configuration options for Gateway API behavior and routing.

- **Bug Fixes**
  - Improved chart-change detection in CI and added automated Helm unit testing.

- **Tests**
[...]

PR #418, opened 2026-07-16
TEMP DIAGNOSTIC — do not review (testing CI trigger in isolation)

drzln:
Diagnostic only — will close immediately once CI trigger status is confirmed. Same content as #417.


drzln:
Diagnostic only — confirmed new PR opens aren't triggering CI right now (repo-side issue), while synchronize on existing PRs works. Closing.

PR #420, opened 2026-07-21
Kubernetes Gateway API on akeyless-gateway — Tier 1 (Gateway + HTTPRoute + redirect)

drzln:
Narrows #399 down to strict `ingress.yaml` parity, per review feedback that the fuller HTTPRoute surface (weighted/canary backends, RequestMirror, rich matches, cross-namespace backendRef + ReferenceGrant) goes beyond what the existing Ingress template does.

Checked the actual `ingress.yaml` template to settle it rather than argue it: it renders exactly one host -> one path -> one backend Service + port. No multiple backends, no weighting, no header/method/query matching, no mirroring, and Kubernetes' core Ingress API can't reference a backend Service in a different namespace at all -- that's a structural limit, not a missing feature. Every one of the richer HTTPRoute capabilities in #399 goes beyond that baseline.

This PR ships only what matches that baseline:
- `Gateway` (optional, `gateway.create`), `HTTPRoute` (simple path match, single backend), and the HTTP->HTTPS redirect route.
- No `matches`/`filters`/multi-`backendRefs`/`weight`/`timeouts` fields, no `ReferenceGrant` template -- none of it renders because none of it exists in this chart yet.

The richer HTTPRoute surface isn't dropped, just deferred -- #399 keeps that content intact as a separate, explicitly-optional follow-up, so it can land iteratively once there's a concrete need for it, instead of shipping now as unused surface area.

Validated:
- `helm unittest`: 22/22 (4 suites; the 2 ReferenceGrant-specific cases moved out with the deferred content)
- `helm lint`: clean
- Real `helm template` render checked by hand -- exactly host+path+single-backend, matching the diff above

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes Gateway API support with configurable Gateway, HTTPRoutes, listeners, TLS termination (Terminate), and automatic HTTPS redirects.
  * Introduced a `gatewayAPI` configuration block for backend service/port selection, routing paths/hostnames, metadata (labels/annotations), infrastructure/addressing, and `parentRefs`.
* **Bug Fixes**
  * Strengthened configuration guardrails and clearer validation for common misconfigurations.
* **Tests**
  * Expanded Helm unittest coverage for Gateway/HTTPRoute rendering, redirect behavior, routing, and validation scenarios.
* **Chores**
  * Bumped the Helm chart version to **3.2.1** and improved the chart test workflow (better output handling and Gateway API CRD installation).


drzln:
@coderabbitai review

Alex-Souslik (commented):
One more thing on the `gatewayApi.validate` / `gateway-api-validate.yaml` setup — I think this chart already has the right tool for this and isn't using it: `values.schema.json`.

Checking the guards against where they're actually needed: **G2** (`gatewayClassName` required when `create=true`) and **G5** (unknown `servicePort`) are already re-checked inline in `gateway.yaml`'s `required` call and `httproute.yaml`'s `portNumber` call respectively, under the same render conditions — so their copies in `gatewayApi.validate` are dead code, just caught first because of file ordering.

That leaves **G1**, **G3**, **G4**, **G6** as the ones that genuinely need to fire regardless of which resource template renders (e.g. G3 fires exactly when `gateway.yaml` *doesn't* render). All four are expressible in `values.schema.json`:
- G1 (mutual exclusion with `gateway.ingress`) and G3 (`parentRefs` required when `create=false`) via `if`/`then`/`not`.
- G4 (TLS mode) doesn't even need a conditional — just restrict the `tls.mode` enum to `["Terminate"]`.
- G6 (listeners override + httpsRedirect) via the same `not` pattern as G1.

G5 can't move (schema enums are static, can't reference `backend.ports[].name` dynamically) — but it's already correctly inline in `httproute.yaml`, so nothing's lost by dropping its copy from the validate helper.

Net: `_gateway-api.tpl`'s `gatewayApi.validate` and `gateway-api-validate.yaml` could go away entirely, with the four schema-expressible guards moving into `values.schema.json` (which today only covers `globalConfig.allowedAccessPermissions` — plenty of room to grow) and getting better UX for free (`helm lint`/`helm install` reject bad values before templating even runs, IDE/editor validation against the schema). Happy to pair on the schema additions if useful.

Nitpick I'll admit is mostly taste: `httproute.yaml` and `httproute-redirect.yaml` could be one file with a leading `---`, the same way `templates/svc.yaml` already renders two conditional Services from a single file. Feel free to tell me to go find a hobby.

drzln:
I really like all kinds of generic schema validation so I'll take a first shot then we can review it together and make any small adjustments

drzln:
Took that first shot — the schema rework landed in 576a88e, and I think it matches your breakdown guard-for-guard.

`gatewayApi.validate` and `gateway-api-validate.yaml` are both gone. G1, G3, G4 and G6 moved into `values.schema.json`; G2 and G5 stayed inline for exactly the reasons you gave (G5 can't be a static enum since it has to reference `backend.ports[].name`, and G2's `required` call already fires under the same render condition).

Verified each one actually rejects:

| Guard | Where | Result |
|---|---|---|
| G1 mutual exclusion with `gateway.ingress` | schema `allOf[0]`, if/then/not | rejected |
| G3 `parentRefs` required when `create=false` | schema `allOf[1]` | `at '/gatewayAPI/parentRefs': minItems: got 0, want 1` |
| G4 TLS mode | schema enum `["Terminate"]`, no conditional | `at '/gatewayAPI/gateway/tls/mode': value must be 'Terminate'` |
| G6 listeners + `httpsRedirect` | schema `allOf[2]` | `at '/gatewayAPI/gateway/listeners': maxItems: got 1, want 0` |
| G2 `gatewayClassName` | inline `required`, gateway.yaml:19 | unchanged |
| G5 unknown `servicePort` | inline `fail`, _gateway-api.tpl:42 | unchanged |

You were right about the UX win too — these fire during `helm template`/`helm lint` before rendering, so bad values never reach a cluster.

Also took the nitpick: `httproute-redirect.yaml` is merged into `httproute.yaml` behind a leading `---`, same shape as `svc.yaml`. Not a hobby-finding exercise, it reads better.

**CI evidence:** https://github.com/akeylesslabs/helm-charts/actions/runs/30304278065

The four `tests/gateway-api-*_test.yaml` suites weren't actually running before this — `helm-unittest` isn't wired into chart CI, and `ct install` didn't cover the feature either since `gatewayAPI` defaults off and no `ci/` fixture enabled it. Both are bound now:

| Gate | Result |
|---|---|
| `ct lint` | pass |
| `helm unittest` — 4 suites, 23 cases | pass |
| Gateway API CRDs (standard v1.2.1) into kind | pass |
| `ct install` — Gateway + HTTPRoute applied | pass |
| `release-scripts-test` | pass |

Added `charts/akeyless-gateway/ci/gateway-api-values.yaml` (renders 1 Gateway + 1 HTTPRoute), plus the helm-unittest steps and `ct list-changed` debug-line filter from #401 and the CRD install that `ct install` needs. The Gateway stays unprogrammed in kind since no controller runs there — the point of that step is CRD schema validation on apply, and `helm --wait` doesn't gate on custom resources.

Since #401's workflow change now lives here, #401 is largely redundant.

Alex-Souslik (approved):

stanislav-akeyless (approved):

.github/workflows/chart_test.yaml
```diff
@@ -73,10 +78,38 @@ jobs:
...
         if: steps.list-changed.outputs.changed == 'true'
         uses: helm/kind-action@ef37e7f390d99f746eb8b610417061a60e82a6cc # v1
 
+      - name: Install Gateway API CRDs
+        if: steps.list-changed.outputs.changed == 'true'
+        # ci/gateway-api-values.yaml makes ct install apply Gateway + HTTPRoute.
+        # kind ships no Gateway API CRDs, so without this the apply fails with
+        # "no matches for kind Gateway". Standard channel matches what the chart
+        # targets (gateway.networking.k8s.io/v1).
+        run: kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/standard-install.yaml
```
omriezra, line 111: this should be only installed when you need gateway-api

charts/akeyless-gateway/ci/gateway-api-values.yaml
```diff
@@ -0,0 +1,15 @@
...
+# renders none of the gateway-api/ templates -- the feature would ship with
+# lint + unit coverage only, never validated against live CRD schemas.
+gatewayAPI:
+  enabled: true
+  gateway:
+    create: true
+    # No controller runs in the kind cluster, so the Gateway stays
+    # unprogrammed. That is fine: helm --wait does not gate on custom
+    # resources, and the point here is CRD schema validation on apply.
+    gatewayClassName: cilium
```
omriezra, line 12: do we have cilium installtion?

PR #424, opened 2026-07-28
ASM-19370 | expose SSH keepalive values on the legacy SRA chart

Hanan-Akeyless:
Implements [ASM-19370](https://akeyless.atlassian.net/browse/ASM-19370), the legacy-chart half of [ASM-19019](https://akeyless.atlassian.net/browse/ASM-19019).

Surfaces the four SSH keepalive environment variables as first-class values on the SSH bastion StatefulSet.

```yaml
sshConfig:
  keepalive: {}
    # clientAliveInterval: 120
    # clientAliveCountMax: 2
    # serverAliveInterval: 120
    # serverAliveCountMax: 2
```

The ticket sketched `sshProxy.keepalive.*`, which does not exist in this chart. `sshConfig.keepalive.*` matches the existing structure.

## Defaults ship unset, deliberately

Each env var is emitted only when its value is set, so the bastion image keeps deciding the defaults and an untouched install renders exactly as it did before. Writing 120 and 2 into `values.yaml` would create copies of numbers that already live in the image and its Go config, across two repos, and any drift would silently change behaviour for every customer who never touched the setting. This matches how `allowedBastionUrls` and `config.CAPublicKey` are already presented.

The presence test is `kindIs "invalid" | not`, not a truthiness check. `0` is a documented setting meaning send no probes, and `{{- if .clientAliveInterval }}` would silently drop exactly that value.

The entries sit ahead of the free-form `sshConfig.env` passthrough, so an explicit user override still wins.

## Terminology note for the reviewer

The values comments here say "client to bastion leg" and "bastion to target leg", where the unified gateway chart (#423) says "gateway". That divergence is intentional. This chart's SSH service selects `app: ssh-bastion` directly, so there is no gateway in the SSH data path, and calling it a gateway leg would point a customer at the wrong hop while debugging an idle-timeout drop. Everything else follows the wording PM asked for.

## Version fields, updated by hand

This chart is deliberately excluded from bot auto-release, so its app-version fields are this PR's responsibility.

| Field | From | To |
| --- | --- | --- |
| chart `version` | 3.4.0 | 3.5.0 |
| `appVersion` | 3.2.0_2.1.3 | 3.3.0_2.1.5 |
| `annotations.ztbVersion` | 3.2.0 | 3.3.0 |
| `annotations.ztpVersion` | 2.1.3 | **2.1.5** |

**The `ztpVersion` bump is unrelated to keepalive and deserves a separate look.** The chart pinned 2.1.3 while production runs 2.1.5, because the release bot's legacy-chart gate means portal releases no longer touch this chart. For anyone running `ztpConfig.enabled: true` without pinning `ztpConfig.image.tag`, merging this moves the portal image 2.1.3 to 2.1.5. `ztpConfig.enabled` defaults to `false`, so most installs are unaffected. Happy to split this into its own PR if you would rather keep this one single-purpose.

## Verification

- `helm lint` passes, with and without keepalive values set.
- With the keys unset, `helm template` output is byte-identical to the pre-change baseline apart from the intended chart-version label. No env vars are added.
- With all four set, all four render, including `serverAliveInterval: 0` rendering as `"0"` rather than being dropped.
- A partial set emits only the keys provided.
- Full-chart render parses as valid YAML (6 documents), and renders `zero-trust-bastion:3.3.0` plus `zero-trust-portal:2.1.5`.
- `ci/sra-ssh-values.yaml` now sets a non-default keepalive, so the chart-testing install path exercises the new env vars rather than only the unset case.

## Notes

- Depends on akeylesslabs/zero-trust-bastion#393, which teaches the image to read these variables. Merge that first. The `ztbVersion` bump here means a default install pulls an image that honours them, so only an explicitly pinned older tag would ignore them, which the README now states by version.
- Changing these values requires a pod restart, stated in the values comments and the README.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ASM-19370]: https://akeyless.atlassian.net/browse/ASM-19370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ASM-19019]: https://akeyless.atlassian.net/browse/ASM-19019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added optional SSH keepalive configuration for client and server connections (intervals and count-max), including the ability to disable probes with an interval of `0`.
  * When enabled, keepalive settings are injected into the running pods and take effect on pod restart (unset fields retain image defaults; older image tags may ignore these settings).
* **Documentation**
  * Documented keepalive probe behavior, defaults shipped with the image, and provided parameter examples/tables.
* **Chores**
  * Bumped Helm chart and application versions/annotations for Secure Remote Access.

### DIFF
--- a/charts/akeyless-secure-remote-access/Chart.yaml
+++ b/charts/akeyless-secure-remote-access/Chart.yaml
@@ -17,10 +17,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 
-version: 3.4.0
-appVersion: 3.2.0_2.1.3
+version: 3.5.0
+appVersion: 3.3.0_2.1.5
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 annotations:
-  ztbVersion: 3.2.0
-  ztpVersion: 2.1.3
+  ztbVersion: 3.3.0
+  ztpVersion: 2.1.5
--- a/charts/akeyless-secure-remote-access/README.md
+++ b/charts/akeyless-secure-remote-access/README.md
@@ -193,6 +193,20 @@ The following table lists the configurable parameters of the SSH Bastion chart a
 | `sshConfig.config.logForwarding.settings`           | Log forwarding configuration                                              | `nil`   |
 | `sshConfig.config.logForwarding.existingSecretName` | Existing secret that will old log forwarding configuration                | `nil`   |
 
+### SSH keepalive parameters
+
+A leg is dropped once `interval x count max` seconds pass with no probe reply. The image ships
+`120` and `2`, so the default window is 240 seconds on each leg. Leave a key unset to keep the
+image default. Set an interval to `0` to send no probes on that leg. Changes take effect on pod
+restart, and an `sshConfig.image.tag` older than `3.3.0` ignores these variables.
+
+| Parameter                                | Description                                                                 | Default |
+|------------------------------------------|-----------------------------------------------------------------------------|---------|
+| `sshConfig.keepalive.clientAliveInterval` | Seconds between bastion probes to the connected client                      | `120`   |
+| `sshConfig.keepalive.clientAliveCountMax` | Unanswered client probes before the bastion disconnects the session         | `2`     |
+| `sshConfig.keepalive.serverAliveInterval` | Seconds between bastion probes to the target host                           | `120`   |
+| `sshConfig.keepalive.serverAliveCountMax` | Unanswered target probes before the bastion gives up on the target          | `2`     |
+
 ### Persistence parameters
 
 | Parameter                       | Description                                                                                            | Default                  |
--- a/charts/akeyless-secure-remote-access/ci/sra-ssh-values.yaml
+++ b/charts/akeyless-secure-remote-access/ci/sra-ssh-values.yaml
@@ -25,6 +25,13 @@ sshConfig:
   enabled: true
   service:
     type: ClusterIP
+  # Non-default keepalive so the install path proves the env vars render and the bastion
+  # still starts. Zero on the target leg also covers the "set but falsy" rendering case.
+  keepalive:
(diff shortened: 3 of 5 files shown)


charts/akeyless-secure-remote-access/Chart.yaml
```diff
@@ -17,10 +17,10 @@ type: application
...
-appVersion: 3.2.0_2.1.3
+version: 3.5.0
+appVersion: 3.3.0_2.1.5
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 annotations:
-  ztbVersion: 3.2.0
-  ztpVersion: 2.1.3
+  ztbVersion: 3.3.0
+  ztpVersion: 2.1.5
```
Hanan-Akeyless, line 26: manual bump since we disabled the auto-bumps from frontend repo publishing new portal

charts/akeyless-secure-remote-access/Chart.yaml
```diff
@@ -17,10 +17,10 @@ type: application
...
-version: 3.4.0
-appVersion: 3.2.0_2.1.3
+version: 3.5.0
+appVersion: 3.3.0_2.1.5
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 annotations:
-  ztbVersion: 3.2.0
-  ztpVersion: 2.1.3
+  ztbVersion: 3.3.0
```
Hanan-Akeyless, line 25: next SRA image candidate, fails chart-test since image isn't published yet

yogevk-akeyless (approved):

stanislav-akeyless (approved):

PR #427, opened 2026-08-04
akeyless-k8s-secrets-injection: user-managed webhook TLS certificate

drzln:
Adds `webhook.tls.existingSecretName` so the injector's webhook certificate can be one you manage. Additive: chart `2.2.0` to `2.3.0`, default behaviour byte-for-byte unchanged.

## Why

`templates/apiservice-webhook.yaml` calls `genCA` / `genSignedCert` on every render. Those are non-deterministic and Helm holds no state at template time, so each render emits a different keypair.

A GitOps controller re-renders on every sync cycle, so with ArgoCD auto-sync and self-heal three things drift continuously:

- the `Secret` holding `tls.crt` / `tls.key`
- the `MutatingWebhookConfiguration` `caBundle`, which has to agree with it
- the deployment's `checksum/config` annotation, a sha256 of the render

The first two produce the reported `tls: bad certificate`, as the API server and the pod fall out of step. The third is not in the ticket but means the webhook pods **roll on every sync**, which likely accounts for symptoms beyond the handshake noise. It also made cert-manager unusable, since the chart overwrote whatever cert-manager issued.

## Change

Set `webhook.tls.existingSecretName` and the chart stops generating: no `genCA`, no `genSignedCert`, and the `Secret` is no longer emitted at all. Both halves matter — skipping generation makes the render deterministic, and not emitting the Secret is what stops a controller owning and rewriting it. The render becomes byte-for-byte stable.

A **non-empty** value flips this, not the presence of the key; `existingSecretName: ""` is falsy and behaves exactly like unset.

The CA then comes from either `webhook.tls.caBundle` (raw PEM, base64-encoded by the chart) or `webhook.tls.caBundleAnnotations` (so cert-manager's ca-injector populates it).

Three render-time guards, because every one of these fails silently and cluster-wide — a webhook the API server cannot validate blocks Pod creation in every selected namespace:

| combination | why it is refused |
| --- | --- |
| `existingSecretName` with neither CA source | nothing tells the API server what to trust |
| `existingSecretName` equal to the chart's own full name | `helm upgrade` prunes the Secret dropped from the manifest — the one the pods mount |
| `caBundleAnnotations` without `existingSecretName` | the chart writes its own generated CA there; an injector would replace it with an unrelated one |

## No server-side change required

The webhook already supports an externally-managed certificate. In `akeyless-kubernetes-webhook-secrets`, `vault-secrets-webhook/cert/cert.go` opens a retry-watch on `TLS_SECRET_NAME` / `TLS_SECRET_NAMESPACE` and hot-swaps through the `GetCertificate` callback, so rotation needs no pod restart. The ClusterRole in `webhook-role.yaml` already grants `get`/`watch` on secrets. The chart was the only blocker.

## Example

```yaml
webhook:
  tls:
    existingSecretName: akeyless-injector-tls
    caBundleAnnotations:
      cert-manager.io/inject-ca-from: "akeyless/akeyless-injector-tls"
```

## Two caveats, documented in the chart README

**Startup ordering.** With an external Secret the pod reads the mounted certificate at startup and exits if absent, so pods crash-loop until the issuer populates it, then recover unaided.

**`caBundleAnnotations` moves the drift rather than removing it.** The render omits `caBundle`, ca-injector writes it into the live object, and a self-healing controller strips it on the next sync. The README carries the ArgoCD `ignoreDifferences` + `RespectIgnoreDifferences=true` and Flux `driftDetection.ignore` exclusions — `ignoreDifferences` alone only changes how the diff is computed, so without the sync option the value still gets removed.

## Verification

```
helm unittest                                     9 passed, 9 total
  suite vs pre-change main                        6 failed (the default-path cases pass, correctly)
helm lint, default / api-key / existing-secret    0 failed
existing-secret render, twice, diffed             identical
default render                                    unchanged
```

Coverage is the nine `helm-unittest` cases and the render diff above — this PR is chart-only and adds nothing under `.github/`. An earlier revision installed the existing-secret path into kind with a minted TLS Secret; that was reverted in `de8b1e9` in favour of proving the path by render, and the description had not caught up (thanks @Alex-Souslik).

`lookup` was deliberately not used to stabilize the generated certificate: it returns empty during `helm template`, which is exactly how ArgoCD renders, so it would not fix the drift it appears to fix. Noted in `AGENTS.md`.

## Scope

Deliberately the smallest thing that answers the ticket, and now the only PR open for it. The follow-on work stays on its branches and is tracked internally rather than as four more review requests:

| branch | what |
| --- | --- |
| `injector-render-determinism-gate` | CI gate asserting the render really is byte-stable |
| `injector-cert-manager-certificate` | the chart owning a cert-manager `Certificate` |
| `injector-cabundle-guards` | rejecting the remaining bad `caBundle` combinations |
| `injector-rbac-namespaced` | narrowing the webhook's RBAC. Breaking, and unrelated to this ticket |

## Summary by CodeRabbit

* **New Features**
  * Added support for externally managed webhook TLS certificates.
  * Added configuration for existing TLS Secrets, CA bundles, and CA-injector annotations.
  * Certificates are generated automatically when no existing Secret is configured.
  * Added cert-manager, ArgoCD, and Flux integration guidance.

* **Bug Fixes**
  * Improved TLS validation and handling of invalid configuration combinations.

* **Documentation**
  * Added guidance on required Secret keys, reserved names, and startup behavior.

* **Tests**
  * Added coverage for generated and user-provided certificates and CA configurations.


drzln:
@coderabbitai full review

drzln:
@coderabbitai full review

drzln:
@coderabbitai all three are already addressed — two in this PR, one in #432 above it.

**`RespectIgnoreDifferences`** — present at README.md:114, inside the ArgoCD example, with the explanation at :124 that `ignoreDifferences` alone only affects diff calculation.

**`fullnameOverride`** — README.md:152 states the reserved name is `RELEASE_NAME-akeyless-secrets-injection` *by default and whatever you set in `fullnameOverride` otherwise*, and :155 records that the render-time guard covers that case too.

**Both CA sources set** — real, and rejected in #432, which stacks on this one: `webhook.tls.caBundle and webhook.tls.caBundleAnnotations are both set. They are two writers for the same clientConfig.caBundle field` — with a `failedTemplate` case asserting it. It is not in this PR because this PR is deliberately the minimal additive change for the ticket; the remaining guards are a separate rung.

Worth noting the review is against a PR that changed shape after your last pass: it was split from 3.0.0 to an additive 2.3.0, the RBAC work moved to #434, and the CI/fixture changes moved to #430 and #434.

charts/akeyless-k8s-secrets-injection/values.yaml
```diff
@@ -77,6 +77,12 @@ mutatingWebhook:
   matchConditions: []
 debug: false
 
+webhook:
```
Alex-Souslik: Nit on placement/naming: this chart already has a convention for "bring your own cert" — `gatewayCert.tlsCertsSecretName` a few lines up — which this doesn't follow (it mirrors `akeyless-api-gateway`'s `cache.tls.existingSecretName` pattern instead, from a different chart). More importantly, `webhook:` sitting right next to the pre-existing `mutatingWebhook:` is an easy mix-up: `mutatingWebhook` already documents itself as the pass-through for this exact webhook object's fields, and the new `webhook.tls.*` config is specifically for that same object's serving cert. Nesting under `mutatingWebhook.tls.*` would avoid adding a second, similarly-named top-level key.

charts/akeyless-k8s-secrets-injection/templates/apiservice-webhook.yaml
```diff
@@ -1,37 +1,67 @@
...
+{{ $caBundleAnnotations := $tls.caBundleAnnotations | default dict }}
+{{ $servingCert := dict }}
+{{ $generatedCA := dict }}
+
+{{- if $existingSecret }}
+  {{- if eq $existingSecret $fullName }}
+    {{- fail (printf "webhook.tls.existingSecretName must not be %q, the name the chart uses for its own generated Secret. The chart stops emitting that Secret when existingSecretName is set, so helm upgrade on an existing release deletes the Secret the webhook pods mount. Give the user-managed Secret a different name." $fullName) }}
+  {{- end }}
+  {{- if and (not $caBundle) (not $caBundleAnnotations) }}
+    {{- fail "webhook.tls.existingSecretName is set, so the chart no longer generates a CA. Set webhook.tls.caBundle to the PEM CA that signed the certificate, or webhook.tls.caBundleAnnotations to let a controller such as cert-manager's ca-injector populate it." }}
```
Alex-Souslik: The reasoning in this message isn't quite right: per the k8s API's own doc comment for `WebhookClientConfig.caBundle` (`kubernetes/api` `admissionregistration/v1/types.go`), "If unspecified, system trust roots on the apiserver are used." So an omitted `caBundle` isn't a broken configuration — the API server just falls back to its system trust store. That's a real, if less common, setup (a cert signed by something already in the apiserver's trust store, or fronted by a mesh sidecar). As a conservative default this guard is reasonable — nearly everyone here has a private CA — but the message overclaims ("nothing tells the API server what to trust" isn't true), and there's currently no escape hatch for the system-trust-roots case. Worth rewording, and maybe worth an explicit opt-out flag rather than a hard block.

Alex-Souslik (commented):
Verified the render logic locally (checked out the branch, ran `helm lint` + `helm unittest`, rendered both paths twice to confirm the determinism claim, and manually tripped all three `fail` guards including the untested `fullnameOverride` case) — the mechanics hold up. One thing that doesn't: the **Verification** section of the PR description still describes CI installing the existing-secret path into a live kind cluster via a per-chart `<name>-tls.conf` fixture. That was added in `832ca20`/`83f8e11` but reverted by the last commit on this branch, `de8b1e9` ("prove the existing-secret path by render, not by a second kind install") — its own message says `This PR is now chart-only: no .github changes, no second ci/ values file, no shell added anywhere`, and `git diff main..HEAD -- .github/ ci/` confirms zero net changes there. Coverage is genuinely just the 9 `helm-unittest` cases, which is fine, but please update the description so reviewers aren't told there's live-cluster coverage that isn't in this diff.

A few inline notes below, plus one non-blocking design suggestion.

charts/akeyless-k8s-secrets-injection/templates/apiservice-webhook.yaml
```diff
@@ -1,37 +1,67 @@
...
+{{ $tls := fromYaml (include "vault-secrets-webhook.tls" .) }}
+{{ $existingSecret := include "vault-secrets-webhook.existingTLSSecretName" . }}
+{{ $caBundle := $tls.caBundle | default "" }}
+{{ $caBundleAnnotations := $tls.caBundleAnnotations | default dict }}
+{{ $servingCert := dict }}
+{{ $generatedCA := dict }}
+
+{{- if $existingSecret }}
+  {{- if eq $existingSecret $fullName }}
+    {{- fail (printf "webhook.tls.existingSecretName must not be %q, the name the chart uses for its own generated Secret. The chart stops emitting that Secret when existingSecretName is set, so helm upgrade on an existing release deletes the Secret the webhook pods mount. Give the user-managed Secret a different name." $fullName) }}
```
Alex-Souslik: This guard is well justified — I confirmed a resource that drops out of the rendered manifest gets pruned on `helm upgrade`, so reusing `fullName` here would delete the very Secret the pods are mounting.

Non-blocking suggestion for a follow-up: rather than guarding against this name specifically, consider giving the chart's own generated Secret a dedicated name (e.g. `{{ fullName }}-generated`) instead of `fullName`. That removes the collision class by construction — no `existingSecretName` value could ever collide with the chart's own resource, so this whole check becomes unnecessary. Two caveats worth flagging if you go this route: (1) it changes the Secret name emitted on the *default* path too, for every existing install, not just this new opt-in one, so it's a bigger blast radius than this PR and would need its own migration note; (2) `helm.sh/resource-policy: keep` looked like it might sidestep this more simply, but ArgoCD maps that annotation to its own `Delete=false`, which per ArgoCD's docs is mainly about protecting a resource when the whole Application is deleted — it's not clearly documented as covering the "resource just disappeared from this sync's render" case (that's `Prune=false`'s job), so I wouldn't rely on it here given this chart's GitOps focus. Just flagging as a nicer long-term shape, not asking for a change in this PR.

charts/akeyless-k8s-secrets-injection/tests/webhook-tls_test.yaml
```diff
@@ -0,0 +1,132 @@
+suite: akeyless-secrets-injection webhook TLS
+release:
+  name: injector
+  namespace: akeyless
+tests:
+  - it: "default path generates the Secret and a caBundle"
+    template: templates/apiservice-webhook.yaml
+    asserts:
+      - hasDocuments:
```
Alex-Souslik: `hasDocuments: count: 1` here (and again at line 63) is vacuous — this manifest is a single `kind: List` document either way, so this assertion passes regardless of how many `items` are inside. The actual check for "Secret present vs. absent" is the `items[0].kind`/`items[1]` assertions below, which are the ones doing real work. Not wrong, just doesn't test what it looks like it tests.

charts/akeyless-k8s-secrets-injection/templates/apiservice-webhook.yaml
```diff
@@ -1,37 +1,67 @@
...
+  {{- end }}
+  {{- if and (not $caBundle) (not $caBundleAnnotations) }}
+    {{- fail "webhook.tls.existingSecretName is set, so the chart no longer generates a CA. Set webhook.tls.caBundle to the PEM CA that signed the certificate, or webhook.tls.caBundleAnnotations to let a controller such as cert-manager's ca-injector populate it." }}
+  {{- end }}
```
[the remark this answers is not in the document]
drzln: Valid, and deliberately not fixed here. `caBundle` and `caBundleAnnotations` set together is one of four bad combinations that render clean today. All four land in one change rather than one at a time, on the `injector-cabundle-guards` branch: both sources set, `caBundle` without `existingSecretName` (silently overwritten by the generated CA), a `caBundle` carrying no PEM block, and the annotations case already guarded here.

This PR stays the minimum that answers the ticket.

charts/akeyless-k8s-secrets-injection/README.md
```diff
@@ -59,6 +59,105 @@ mutatingWebhook:
...
+server unable to validate the webhook. Exclude the field from drift detection. ArgoCD:
+
+```yaml
+syncPolicy:
+  syncOptions:
+    - RespectIgnoreDifferences=true
+ignoreDifferences:
+  - group: admissionregistration.k8s.io
+    kind: MutatingWebhookConfiguration
+    name: RELEASE_NAME-akeyless-secrets-injection
```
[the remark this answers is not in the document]
drzln: Already the case. Both examples use `RELEASE_NAME-akeyless-secrets-injection`, the rendered default: the `dnsNames` entry in the cert-manager `Certificate` and the `name` in `ignoreDifferences`. The `fullnameOverride` case is documented under "the Secret must not be named after the chart's own generated full name", which states that the reserved name follows `fullnameOverride` and that the chart rejects it at render time, including that case.

No change needed unless you want the caveat repeated at each example rather than stated once.

drzln:
All four addressed in d8190b8, naming included.

charts/akeyless-k8s-secrets-injection/tests/webhook-tls_test.yaml
```diff
@@ -0,0 +1,145 @@
...
+  - it: "tls never reaches the rendered webhook object"
+    template: templates/apiservice-webhook.yaml
+    set:
+      mutatingWebhook.tls.existingSecretName: akeyless-injector-tls
+      mutatingWebhook.tls.caBundle: TESTCA
+    asserts:
+      - isNull:
+          path: items[0].webhooks[0].tls
+      - isNull:
+          path: items[0].tls
```
Alex-Souslik: The `items[0].webhooks[0].tls` check above is the one doing real work. This second assertion (`items[0].tls`) is vacuous the same way the old `hasDocuments: count: 1` was: `items[0]` is the whole `MutatingWebhookConfiguration`, which never had a top-level `.tls` field before or after this fix, so it can't fail regardless of whether the omit-list change works. Harmless, not blocking — just flagging so it doesn't read as covering more than it does.

Alex-Souslik (approved):
Re-verified locally (checked out the branch, ran `helm lint` + `helm unittest` — 10/10 — and rendered the existing-secret path myself). All four prior comments are genuinely addressed, including one you caught that I didn't: moving `tls` under `mutatingWebhook` meant it would've been splatted into the rendered webhook object by the existing `omit $mutatingWebhook ...` pass-through, and you added it to the omit list plus a test for it. Confirmed by rendering that no `tls:` key leaks into `webhooks[0]`. One tiny nit inline, not blocking — approving.

Alex-Souslik (approved):

stanislav-akeyless (approved):

PR #430, opened 2026-08-06
gate render determinism for the ci paths that must be GitOps-stable

drzln:
Stacks on #427. Adds the CI assertion that #427's premise rests on, which nothing was checking.

## Why

#427 exists so that two renders of the existing-secret path produce identical YAML. No check asserted that. The unittest cases pin the shape of a single render, so a reintroduced `genCA`, `randAlphaNum`, `uuidv4` or `now` call passes all of them and the drift is back with the suite green.

An adversarial pass over the gate produced regressions that survive it today. Two were confirmed by applying them in an isolated worktree and re-running the whole gate:

| regression | gate result |
| --- | --- |
| a `now`-valued annotation on the deployment pod template | 16/16 pass, lint clean, render nondeterministic |
| a per-render annotation on the `MutatingWebhookConfiguration`, which also flows into `checksum/config` | same |

Neither touches a certificate. Both reproduce the pod-roll half of the original symptom.

## Change

`.github/chart-fixtures/<chart>-deterministic.conf` lists the chart's `ci/` values files whose render must be byte-identical. `tools/render-determinism.tlisp` renders each twice and compares. The injector lists `existing-secret-values.yaml`.

Opt-in, so it is available to every chart in the repo without changing any of them, and a chart with no such file is skipped. It runs before the kind cluster is created, so the cheap check fails first.

The default path is deliberately not listed. It generates a keypair per render by design, which is the defect the rest of this stack addresses.

## Why a script and not a `run:` block

The first cut was 26 lines of bash inline in the workflow, doing real work: reading a sidecar conf, skipping comments, looping, rendering, diffing, choosing an exit code. That cannot be linted, cannot be run locally without recreating the job environment, and has nowhere to put a test. The logic now lives in `tools/render-determinism.tlisp` and the job is three `uses:` steps and one `with:`, with no `run:` in it:

```yaml
  render-determinism:
    runs-on: ubuntu-24.04
    steps:
      - uses: actions/checkout@de0fac2
      - uses: azure/setup-helm@dda3372
      - uses: pleme-io/actions/tatara-script@95c9eb1
        with:
          script-file: tools/render-determinism.tlisp
```

It is a **separate job** because the `tatara-script` release binary links against glibc 2.39 and `chart-test` is pinned to `ubuntu-22.04` (glibc 2.35). That pin is deliberate, so the check moved to its own 24.04 job rather than weakening it. The check renders only, needs no cluster, and now fails independently of the kind job instead of blocking it.

The script derives its work from the conf files themselves rather than from `CHANGED_CHARTS`, so the job needs neither `ct` nor Python, and every opted-in chart is checked on every run. Two renders are cheap and the invariant is not per-change: a chart that has declared itself GitOps-stable must stay that way whatever moved.

Three things the port fixed rather than translated:

- `helm`'s exit status is checked explicitly instead of relying on the job shell's `-e`, and its stderr is surfaced with the file annotation
- the comparison is on captured stdout rather than two temp files and a `diff`
- a difference is reported as the first differing line from each render, which names the offending field directly

Run it yourself, no arguments and no environment:

```
tatara-script tools/render-determinism.tlisp
tatara-script lint tools/render-determinism.tlisp
```

## Verification

Local, against the real chart with real helm:

```
gate on this branch                              two renders identical, exit 0
randAlphaNum annotation on the deployment        exit 1, prints both values
conf entry naming a missing values file          exit 1, annotates the conf
conf naming a chart that does not exist          exit 1, annotates the conf
helm template itself failing                     exit 1, surfaces helm's own error
tatara-script lint                               0 violations
actionlint chart_test.yaml                       clean
helm unittest                                    16 passed, 16 total
```

`tatara-script lint` is what rules out an invented builtin, which is the failure mode that matters when the logic moves language.

No chart file changes, so no version bump.

## Summary by CodeRabbit

* **Tests**
  * Added automated checks to verify that Helm chart rendering produces consistent results across repeated runs.
  * Improved chart test workflows with pinned tooling and actions for more reliable validation.
  * Added a deterministic chart fixture for Kubernetes secrets injection.

* **Documentation**
  * Added guidance for maintaining deterministic Helm templates and fixtures.


drzln:
Closing this in favour of tracking it internally. The branch `injector-render-determinism-gate` stays as-is and the work continues there; only #427 stays open as a PR, so the review queue holds one request rather than five. Reopening later is a click if we'd rather review it here.

PR #431, opened 2026-08-06
let the chart own a cert-manager Certificate for the webhook

drzln:
Stacks on #430. Closes the half of the problem `existingSecretName` leaves open.

## Why

`existingSecretName` stops the drift, but only for someone who already has a Secret and knows to wire three values together. Leave it unset, which every existing install does, and the chart still calls `genCA`/`genSignedCert` on every render. Confirmed on two consecutive default renders:

```
tls.crt  tls.key  ca.crt  caBundle  checksum/config     all five change
```

So every `helm upgrade`, including one that changes nothing else, rotates the webhook certificate and rolls the pods. Cert and caBundle rotate together in one apply, so this is bounded rather than continuous, but with `failurePolicy: Fail` the rollout window blocks Pod creation in every selected namespace.

Where cert-manager is present the chart can close this itself.

## Change

`webhook.tls.certManager.enabled` renders a `Certificate`, resolves the chart's effective TLS Secret name to the one cert-manager issues, and defaults the CA-injection annotation to that `Certificate`. One flag replaces the three-part manual wiring:

```yaml
webhook:
  tls:
    certManager:
      enabled: true
      issuerRef:
        name: my-issuer
        kind: Issuer
```

The render stays deterministic, since a `Certificate` is a static document. `certificateName` and `secretName` are independent, `dnsNames` defaults to the webhook Service's in-cluster names, and `duration`, `renewBefore` and `privateKey` pass through untouched.

Two guards: an empty `issuerRef.name`, and `certManager` together with `existingSecretName`, which are alternatives rather than layers.

`lookup` is still not the answer here and is not used. It returns empty during `helm template`, which is exactly how ArgoCD renders, so it would fix the CLI path, leave the GitOps path broken, and make the chart's behaviour depend on whether a cluster connection happens to exist.

## Verification

```
helm unittest                                   27 passed, 27 total
positive control, Certificate template removed  7 failed
helm lint  default / api-key / existing-secret / cert-manager   0 failed
cert-manager path rendered twice, diffed        identical
default render                                  no Certificate, output unchanged
```

Chart `3.0.0` to `3.1.0`. Additive; default behaviour untouched.

## Not covered

No `ci/` values file for this path, so the kind layer does not exercise it. cert-manager is not installed in that cluster, the `Certificate` would never be issued and the pod would crash-loop waiting for the Secret. Installing cert-manager in kind and proving a real issued certificate is its own layer.

## Summary by CodeRabbit

* **New Features**
  * Added optional cert-manager integration for automatic webhook TLS certificate creation and rotation.
  * Added configurable certificate names, DNS names, duration, renewal, private-key settings, and issuer references.
  * Added automatic CA bundle injection when no explicit CA configuration is provided.
  * Updated chart documentation with cert-manager setup and manual Secret guidance.
  * Updated the Helm chart version to 2.4.0.

* **Tests**
  * Added coverage for certificate rendering, customization, CA injection, Secret mounting, and invalid configurations.

### DIFF
--- a/charts/akeyless-k8s-secrets-injection/Chart.yaml
+++ b/charts/akeyless-k8s-secrets-injection/Chart.yaml
@@ -8,7 +8,7 @@ maintainers:
 type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.3.0
+version: 2.4.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: 0.39.0
--- a/charts/akeyless-k8s-secrets-injection/README.md
+++ b/charts/akeyless-k8s-secrets-injection/README.md
@@ -78,7 +78,34 @@ neither is set. Both belong to this mode only: rendering also fails if `caBundle
 without `existingSecretName`, since the chart writes its own generated CA into `caBundle` there and
 an injector would overwrite it with an unrelated one.
 
-With cert-manager issuing the certificate and injecting the CA bundle:
+#### With cert-manager: let the chart own the Certificate
+
+If cert-manager is installed, this is the shortest correct configuration. The chart renders the
+`Certificate`, points itself at the Secret cert-manager issues, and annotates the
+`MutatingWebhookConfiguration` so ca-injector fills in `caBundle`. No Secret to create by hand:
+
+```yaml
+webhook:
+  tls:
+    certManager:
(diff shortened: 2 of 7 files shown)


drzln:
Closing this in favour of tracking it internally. The branch `injector-cert-manager-certificate` stays as-is and the work continues there; only #427 stays open as a PR, so the review queue holds one request rather than five. Reopening later is a click if we'd rather review it here.

PR #432, opened 2026-08-06
reject the caBundle combinations that silently produce a broken webhook

drzln:
Stacks on #431. Turns four values combinations that render happily and leave the cluster wrong into render failures.

## Why

Each of these produces a webhook the API server cannot validate, or silently discards what the operator asked for. All four passed the gate before this change.

| combination | what happens today |
| --- | --- |
| `caBundle` and `caBundleAnnotations` together | two writers for one field: the chart renders the literal, an injector overwrites it, a self-healing controller reverts that, and validation fails intermittently |
| `caBundle` without `existingSecretName` | overwritten by the generated CA on the next line, so the pinned CA is silently discarded |
| a `caBundle` carrying no PEM block | base64-encoded into `clientConfig` unparsed; an unparsable `caBundle` blocks Pod creation in every selected namespace |
| `caBundleAnnotations` without `existingSecretName` | already rejected in #427; the `caBundle` case above is the same shape and was not |

## Change

Three `fail` guards, placed before the existing ones so the cheaper check reports first. The failure mode in every case is silent and cluster-wide, which is why these are render errors rather than warnings.

**Behaviour change.** `caBundle` without `existingSecretName` used to be ignored, and a test pinned that. It now fails, and that test is replaced. Nobody can be relying on it: the value was introduced in #427 and has never been released.

Test values move to a real PEM block so the shape check is exercised rather than worked around.

## Verification

```
helm unittest                          29 passed, 29 total
positive control, guards stripped      3 failed
helm lint  default / api-key / existing-secret    0 failed
yamllint, ct config                    clean
determinism gate                       two renders identical
```

Chart `3.1.0` to `3.2.0`.

## Summary by CodeRabbit

* **New Features**
  * Added validation for webhook TLS settings, including required certificate formats and compatible CA source configurations.
  * Helm rendering now provides clear failures for invalid or conflicting TLS options.

* **Bug Fixes**
  * Improved handling of PEM-formatted CA certificates and existing TLS secrets.

* **Documentation**
  * Updated TLS configuration guidance to reflect the new validation requirements.

* **Tests**
  * Expanded coverage for valid PEM certificates and invalid TLS configurations.

* **Chores**
  * Updated the Helm chart version to 2.5.0.

### DIFF
--- a/charts/akeyless-k8s-secrets-injection/Chart.yaml
+++ b/charts/akeyless-k8s-secrets-injection/Chart.yaml
@@ -8,7 +8,7 @@ maintainers:
 type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.0
+version: 2.5.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: 0.39.0
--- a/charts/akeyless-k8s-secrets-injection/README.md
+++ b/charts/akeyless-k8s-secrets-injection/README.md
@@ -74,9 +74,11 @@ is rotated, so renewal does not require a pod restart.
 
 Because the chart no longer owns a CA in this mode, you must also tell the API server which CA to
 trust, using either `webhook.tls.caBundle` or `webhook.tls.caBundleAnnotations`. Rendering fails if
-neither is set. Both belong to this mode only: rendering also fails if `caBundleAnnotations` is set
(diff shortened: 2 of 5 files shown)


drzln:
Closing this in favour of tracking it internally. The branch `injector-cabundle-guards` stays as-is and the work continues there; only #427 stays open as a PR, so the review queue holds one request rather than five. Reopening later is a click if we'd rather review it here.

PR #434, opened 2026-08-06
akeyless-k8s-secrets-injection: narrow the webhook's secret access to its own namespace

drzln:
Top of the stack, and the only breaking change in it. Chart `2.5.0` to `3.0.0`.

**Not part of.** It was found while adding that ticket's test coverage and was originally bundled into #427; it is split out here so the customer's feature is an additive `2.3.0` and this gets judged on its own risk.

## The bug

The `ClusterRole` and `ClusterRoleBinding` had fixed names shared by every release, so a second install of this chart in any cluster fails on Helm ownership. **A customer cannot run the injector in two namespaces today.**

## Why the fix is not a rename

The `ClusterRole` existed to grant `get`/`watch` on secrets cluster-wide, and the webhook never used it:

| evidence | result |
| --- | --- |
| `cert.go:95` | `Secrets(cm.tlsSecretNamespace).Watch(...)` — its own namespace |
| controller-runtime `List` calls | `NamespaceList`, `DeploymentList`, `DaemonSetList`, `StatefulSetList` — no `SecretList` |
| `GetListItemsForRestart` | the Akeyless API, not the cluster, despite `listSecretChanged` calling it |
| the webhook repo's own copy of this chart | `ClusterRole` gated on `restartRollout`, and carries no `secrets` rule |

So the grant moves to a namespaced `Role`/`RoleBinding` in the release namespace, and the `ClusterRole` remains only for `restartRollout`, which genuinely reads across the cluster. A default install now creates **no cluster-scoped RBAC** — the collision is removed by construction rather than worked around. The chart still creates a cluster-scoped `MutatingWebhookConfiguration`; that is unrelated and unchanged.

## Breaking

`akyeless-injector-role` and `akeyless-injector-role-binding` are gone. `helm upgrade` removes them. The README has an "Upgrading to 3.0.0" section naming both, for anyone with an external RoleBinding, policy or aggregation rule pointing at either.

## Verification

```
helm unittest                                  29 passed, 29 total  (7 new RBAC cases)
  RBAC suite vs pre-change main                7 failed, correctly
helm lint                                      0 failed
default install, cluster-scoped RBAC objects   0
```

Seven cases pin the shape, including that the `restartRollout` `ClusterRole` never carries the secrets grant, and that a second release in the same cluster gets its own names.

### DIFF
--- a/.github/chart-fixtures/akeyless-k8s-secrets-injection-tls.conf
+++ b/.github/chart-fixtures/akeyless-k8s-secrets-injection-tls.conf
@@ -1,0 +1,1 @@
+injector-tls-ci *.chart-testing.svc
--- a/.github/workflows/chart_test.yaml
+++ b/.github/workflows/chart_test.yaml
@@ -143,6 +143,25 @@ jobs:
             if [[ -d "$fixture_dir" ]]; then
               kubectl apply -f "$fixture_dir/" -n chart-testing
             fi
+            tls_conf=".github/chart-fixtures/${chart_name}-tls.conf"
+            if [[ -f "$tls_conf" ]]; then
+              while read -r secret_name dns_name _rest; do
+                [[ -z "${secret_name}" || "${secret_name}" == \#* ]] && continue
+                openssl req -x509 -newkey rsa:2048 -nodes -days 3650 \
+                  -subj "/CN=${dns_name}" \
+                  -addext "subjectAltName=DNS:${dns_name}" \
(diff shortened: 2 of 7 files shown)


drzln:
Closing this in favour of tracking it internally. The branch `injector-rbac-namespaced` stays as-is and the work continues there; only #427 stays open as a PR, so the review queue holds one request rather than five. Reopening later is a click if we'd rather review it here.

